PR #15, opened 2021-08-31
Added extended installation script. [API-694]

oztasozgurcan:


yuce:
@oztasozgurcan Can't we upload the executable binary instead of a compressed archive? The installation process would be more straightforward if we don't have to extract an archive.

yuce (approved):

PR #18, opened 2021-09-03
Improved installation scripts' user experience. [API-694]

oztasozgurcan:
I've recreated a set of installation scripts which provides better user experience than the previous versions.


yuce:
@oztasozgurcan Instead of being minimalistic, it seems like everything gets even more complicated with this PR.

PR #19, opened 2021-09-23
change default config file path

utku-caglayan:
change default config path from "~/.hzc.yaml" to ".local/share/hz-cli/.hzc.yaml"

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -23,9 +23,8 @@ import (
 
 	clusterDir "github.com/hazelcast/hazelcast-commandline-client/commands/cluster"
 	mapDir "github.com/hazelcast/hazelcast-commandline-client/commands/types/map"
+	"github.com/hazelcast/hazelcast-commandline-client/internal"
 )
-
-const DefaultConfigFilePath = ".local/share/hz-cli/.hzc.yaml"
 
 var (
 	cfgFile   string
@@ -49,7 +48,7 @@ func Execute() {
 }
 
 func decorateRootCommand(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", fmt.Sprintf("config file (default is $HOME/%s)", DefaultConfigFilePath))
+	cmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", fmt.Sprintf("config file (default is %s)", internal.DefaultConfigFile()))
 	cmd.PersistentFlags().StringVarP(&addresses, "address", "a", "", "addresses of the instances in the cluster.")
 	cmd.PersistentFlags().StringVarP(&cluster, "cluster-name", "n", "", "name of the cluster that contains the instances.")
 	cmd.PersistentFlags().StringVar(&token, "cloud-token", "", "your Hazelcast Cloud token.")
--- a/commands/root.go
+++ b/commands/root.go
@@ -19,12 +19,13 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/spf13/cobra"
+
 	clusterDir "github.com/hazelcast/hazelcast-commandline-client/commands/cluster"
 	mapDir "github.com/hazelcast/hazelcast-commandline-client/commands/types/map"
-	"github.com/spf13/cobra"
 )
 
-const DefaultConfigFile = ".hzc.yaml"
+const DefaultConfigFilePath = ".local/share/hz-cli/.hzc.yaml"
 
 var (
 	cfgFile   string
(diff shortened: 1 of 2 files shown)


internal/config.go
```diff
@@ -31,6 +31,7 @@ import (
 )
 
 const defaultConfigFilename string = ".hzc.yaml"
```
yuce, line 33: Let's use `config.yaml` as the default configuration file name.

commands/root.go
```diff
@@ -19,12 +19,13 @@ import (
...
 
+	"github.com/spf13/cobra"
+
 	clusterDir "github.com/hazelcast/hazelcast-commandline-client/commands/cluster"
 	mapDir "github.com/hazelcast/hazelcast-commandline-client/commands/types/map"
-	"github.com/spf13/cobra"
 )
 
-const DefaultConfigFile = ".hzc.yaml"
+const DefaultConfigFilePath = ".local/share/hz-cli/.hzc.yaml"
```
yuce, line 28: Let's remove this constant and have a `homeDir(path)` function which would return something like:
```
filepath.Join(os.UserHomeDir(), ".local/share/hz-cli". path)
```

commands/root.go
```diff
@@ -48,7 +49,7 @@ func Execute() {
 }
 
 func decorateRootCommand(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", fmt.Sprintf("config file (default is $HOME/%s)", DefaultConfigFile))
+	cmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", fmt.Sprintf("config file (default is $HOME/%s)", DefaultConfigFilePath))
```
yuce, line 52: Let's use the `homeDir` function here, like: `homeDir(defaultConfigFilename)`

internal/config.go
```diff
@@ -31,6 +31,7 @@ import (
...
+const defaultConfigPathFromHomeDir string = ".local/share/hz-cli"
```
yuce, line 34: Let's remove this constant

yuce (approved):
Thanks, here are a few comments to make the PR better.

yuce (changes requested):
Just a few comments to make the PR better.

commands/root.go
utku-caglayan: It seems we do not use config path for anything else. 

`func DefaultConfigFile() string {
	homeDirectoryPath, _ := os.UserHomeDir()
	return filepath.Join(homeDirectoryPath, DefaultConfigPathFromHomeDir, DefaultConfigFilename)
}`

What do you think for this instead of "homedir" func?

yuce: If we have the `homeDir` function, `DefaultConfigPath` function could be written as:
```
func DefaultConfigPath() string {
    return homeDir("config.yaml")
}
```
But if you think we won't have anything other than the configuration file in the home directory, it is OK to only have the `DefaultConfigPath` function. We may later add `homeDir` if necessary.

utku-caglayan: Sounds great, thanks

internal/config.go
```diff
@@ -45,10 +44,16 @@ func registerConfig(config *hazelcast.Config, confPath string) error {
 	if out, err = yaml.Marshal(config); err != nil {
 		return err
 	}
+
+	filePath, _ := filepath.Split(confPath)
+	if err = os.MkdirAll(filePath, os.ModePerm); err != nil {
+		return fmt.Errorf("cannot create parent directories for configuration file: %w", err)
+	}
+
 	if err = ioutil.WriteFile(confPath, out, 0666); err != nil {
```
yuce, line 53: This is not your code, but can you change `0666` to `0600`? So only the user has r/w access to it?

internal/config.go
```diff
@@ -123,3 +126,8 @@ func MakeConfig(cmd *cobra.Command) (*hazelcast.Config, error) {
 	}
 	return config, nil
 }
+
+func DefaultConfigFile() string {
+	homeDirectoryPath, _ := os.UserHomeDir()
```
yuce, line 131: How about renaming that to `home`?

yuce: Also, the error should not be ignored. Ideally we should return the error, but if you think it is hard to integrate with the rest of the code you can also panic with something like:
```
panic(fmt.Error("retrieving home directory: %w", err))
```

internal/config.go
```diff
@@ -123,3 +126,8 @@ func MakeConfig(cmd *cobra.Command) (*hazelcast.Config, error) {
 	}
 	return config, nil
 }
+
+func DefaultConfigFile() string {
```
yuce, line 130: How about renaming the function to `DefautConfigPath`?

commands/root.go
```diff
@@ -19,13 +19,13 @@ import (
 	"fmt"
 	"log"
 
+	"github.com/spf13/cobra"
+
 	clusterDir "github.com/hazelcast/hazelcast-commandline-client/commands/cluster"
 	mapDir "github.com/hazelcast/hazelcast-commandline-client/commands/types/map"
```
yuce, line 25: This is not your code, but can you change `clusterDir` to something like `clustercmd` and `mapDir` to `mapcmd`?

yuce (changes requested):
Thanks, just a few changes...

internal/config.go
utku-caglayan: When I examined os.UserHomeDir, I saw that it only returns error if it is undefined, which very unlikely in any modern system. Returning error will make it difficult to use this function. If we must, I think panic is a better choice for this

commands/root.go
utku-caglayan: Sorry, handled this but somehow reverted I guess, will do

yuce (approved):
Thanks, looks good!

PR #23, opened 2021-10-08
Refactor errors for hazelcast version 5

utku-caglayan:


commands/cluster/cluster.go
```diff
@@ -34,3 +39,12 @@ func init() {
 	ClusterCmd.AddCommand(clusterShutdownCmd)
 	ClusterCmd.AddCommand(clusterVersionCmd)
 }
+
+func isCloudInvocation(cmd *cobra.Command) bool {
+	token, _ := cmd.InheritedFlags().GetString("cloud-token")
+	if token != "" {
+		fmt.Println("Hazelcast clc does not support cluster operations on cloud. Go checkout https://github.com/hazelcast/hazelcast-cloud-cli for cluster management on cloud.")
```
yuce, line 46: How about: "Cluster operations on Cloud are not supported. Checkout https://github.com/hazelcast/hazelcast-cloud-cli for cluster management on Cloud."

yuce: Note that this function would be more valuable with just returning whether a command is called with Cloud parameters. So, the caller of this function can decide what to do (print a warning in this case).

The function then could be simplifed to `return token != ""`.

yuce: If you'd like to keep the side effect, it would be better to rename the function to something like: `ensureNotCloud`.

commands/types/map/map.go
```diff
@@ -65,17 +66,19 @@ func getMap(clientConfig *hazelcast.Config, mapName string) (result *hazelcast.M
 		clientConfig = &hazelcast.Config{}
 	}
 	clientConfig.Cluster.ConnectionStrategy.Retry.InitialBackoff = types.Duration(1 * time.Second)
-	hzcClient, err := hazelcast.StartNewClientWithConfig(ctx, *clientConfig)
+	configCopy := *clientConfig
```
yuce, line 69: It's better to use `clientConfig.Clone()`.

internal/errors.go
```diff
@@ -32,48 +32,61 @@ func ErrorRecover() {
...
+	if errors.Is(err, syscall.ECONNRESET) {
+		return "Cannot access Hazelcast REST API. Is it enabled? Check this link to find out more: https://docs.hazelcast.com/hazelcast/latest/maintain-cluster/rest-api#enabling-rest-api", true
 	}
 	return "", false
 }
 
-func TranslateNetworkError(err error) (string, bool) {
-	var netErr net.Error
+func TranslateNetworkError(err error, isCloudCluster bool) (string, bool) {
+	cannotConnectErr := "Can not connect to Hazelcast Cluster. Please make sure Hazelcast cluster is reachable, up and running. Check this link to create a local hazelcast cluster: https://docs.hazelcast.com/hazelcast/latest/getting-started/quickstart"
```
yuce, line 59: `Make sure` is enough, no need for `Please`.

yuce: Let's use capitalized `hazelcast`

internal/errors.go
```diff
@@ -32,48 +32,61 @@ func ErrorRecover() {
...
+	/*var netErr net.Error
 	if errors.As(err, &netErr) {
 		if netErr.Timeout() {
-			return "Can not connect to Hazelcast Cluster. Please make sure Hazelcast cluster is reachable, up and running. Check this link to create a IMDG cluster: https://docs.hazelcast.com/imdg/latest/getting-started.html", true
+			return cannotConnectErr, true
 		}
 		var addrErr *net.AddrError
 		if errors.As(err, &addrErr) {
-			return "Invalid cluster address. Please make sure Hazelcast cluster is reachable, up and running. Check this link to create a IMDG cluster: https://docs.hazelcast.com/imdg/latest/getting-started.html", true
+			return "Invalid cluster address. Please make sure Hazelcast cluster is reachable, up and running. Check this link to create a local hazelcast cluster: https://docs.hazelcast.com/hazelcast/latest/getting-started/quickstart", true
```
yuce, line 67: I guess you forgot to delete this commented code.

internal/errors.go
```diff
@@ -32,48 +32,61 @@ func ErrorRecover() {
...
 	var syscallErr syscall.Errno
 	//TODO these syscall errors seem platform specific, decide on what to do
 	if errors.As(err, &syscallErr) && syscallErr == syscall.ECONNREFUSED {
-		return "Can not connect to Hazelcast Cluster. Please make sure Hazelcast cluster is reachable, up and running. Check this link to create a IMDG cluster: https://docs.hazelcast.com/imdg/latest/getting-started.html", true
+		return cannotConnectErr, true
 	}
 	if errors.Is(err, context.DeadlineExceeded) {
-		return "Can not connect to Hazelcast Cluster. Please make sure Hazelcast cluster is reachable, up and running. Check this link to create a IMDG cluster: https://docs.hazelcast.com/imdg/latest/getting-started.html", true
+		if isCloudCluster {
+			return "Can not connect to Hazelcast Cloud Cluster. Please make sure that provided cloud-token and cluster-name parameters are correct.", true
```
yuce, line 83: No need for `Please`.

internal/errors.go
```diff
@@ -32,48 +32,61 @@ func ErrorRecover() {
...
+		}
+		return cannotConnectErr, true
+	}
+	var addrErr *net.AddrError
+	if errors.As(err, &addrErr) {
+		return "Invalid cluster address. Please make sure Hazelcast cluster is reachable, up and running. Check this link to create a local hazelcast cluster: https://docs.hazelcast.com/hazelcast/latest/getting-started/quickstart", true
```
yuce, line 89: No need for `Please`. Let's capitalize `hazelcast`.

yuce (changes requested):

commands/cluster/cluster.go
utku-caglayan: I hesitated while doing this. Side effect removes log line duplication but I think it was not the right way to go. Lets do as you suggested. I will fix this

commands/types/map/map.go
utku-caglayan: Thank you, fixed

commands/cluster/cluster.go
```diff
@@ -27,6 +27,7 @@ var ClusterCmd = &cobra.Command{
 	Long:  `administrative cluster operations which controls a Hazelcast cluster by manipulating its state and other features`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if isCloudInvocation(cmd) {
+			fmt.Println("Cluster operations on cloud are not supported. Checkout https://github.com/hazelcast/hazelcast-cloud-cli for cluster management on cloud.")
```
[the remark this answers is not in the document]
utku-caglayan: Thanks 👍

PR #25, opened 2021-10-13
introduce env variable to call completion script from make file with ease

utku-caglayan:


internal/connect.go
```diff
@@ -24,18 +25,25 @@ import (
...
 type RESTCall struct {
 	url    string
 	params string
 }
 
 func CallClusterOperation(config *hazelcast.Config, operation string, state *string) (*string, error) {
-	obj := NewRESTCall(config, operation, state)
+	obj, err := NewRESTCall(config, operation, *state)
+	if err != nil {
+		if err == InvalidStateErr {
```
yuce, line 38: Can we use `errors.Is(err, InvalidStateErr)` here?

commands/root.go
```diff
@@ -53,6 +55,10 @@ func decorateRootCommand(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVarP(&cluster, "cluster-name", "n", "", "name of the cluster that contains the instances.")
 	cmd.PersistentFlags().StringVar(&token, "cloud-token", "", "your Hazelcast Cloud token.")
 	cmd.CompletionOptions.DisableDefaultCmd = true
+
+	if mode := os.Getenv("MODE"); strings.EqualFold(mode, "dev") { // This is used to generate completion scripts
```
yuce, line 59: It's OK to use env variables for this for the prototype, but we may consider something like build tags later.

yuce (approved):
Looks good, just a comment about error handling.

commands/root.go
utku-caglayan: Sounds good 👍

yuce: Is there a reason we consider `MODE` to be case insensitive?

internal/connect.go
```diff
@@ -24,18 +25,27 @@ import (
...
 	url    string
 	params string
 }
 
 func CallClusterOperation(config *hazelcast.Config, operation string, state *string) (*string, error) {
-	obj := NewRESTCall(config, operation, state)
+	obj, err := NewRESTCall(config, operation, *state)
+	if err != nil {
+		if errors.Is(err, InvalidStateErr) {
+			fmt.Println("Error: invalid new state. It should be one the following:\n", ClusterStateActive, ClusterStateFrozen, ClusterStateNoMigration, ClusterStatePassive)
```
yuce, line 39: `\n` won't work with `fmt.Println`.

scripts/install.sh
```diff
@@ -40,14 +40,14 @@ echo
 read -rd '' addToPathDirectivesZSH << EOF
 * Add \$HOME/.local/bin to PATH to access hzc from any directory
   Append the line below in your .zshrc file if it doesn't exists
-  export PATH=\$HOME/.local/bin:\$PATH" >> \$HOME/.zshrc
+  export PATH=\$HOME/.local/bin:\$PATH >> \$HOME/.zshrc
```
yuce, line 43: The message "Append the line below in your .zshrc file if it doesn't exists" is a bit confusing. The user may add `export PATH=\$HOME/.local/bin:\$PATH >> \$HOME/.zshrc` to their zshrc instead of running that line.

scripts/install.sh
```diff
@@ -40,14 +40,14 @@ echo
...
 
 EOF
 
 read -rd '' addToPathDirectivesBASH << EOF
 * Add \$HOME/.local/bin to PATH to access hzc from any directory
   Append the line below in your .bashrc file if it doesn't exists
-  export PATH=\$HOME/.local/bin:\$PATH" >> \$HOME/.bashrc
+  export PATH=\$HOME/.local/bin:\$PATH >> \$HOME/.bashrc
```
yuce, line 50: The same comment.

internal/connect.go
utku-caglayan: It looked okay to me but I can move it to the same line

scripts/uninstall.sh
```diff
@@ -0,0 +1,33 @@
...
+if [[ -r $HOME/.zshrc ]]; then
+    echo "You can remove ~/.local/bin from your path by removing a line in the form \"PATH=\$HOME/.local/bin:\$PATH\" from ~/.zshrc"
+    echo
+fi
+if [[ -r $HOME/.bashrc ]]; then
+    echo "You can remove ~/.local/bin from your path by removing a line in the form \"PATH=\$HOME/.local/bin:\$PATH\" from ~/.bashrc"
+    echo
+fi
+
+rm -rf ${HZCLI_HOME} #remove hzcli related files (config.yaml,.hzc.yaml,autocompletion scripts)
```
yuce, line 16: This is OK for this prototype, but in the next one we should consider keeping the configuration in case the user modified it.

commands/root.go
utku-caglayan: I thought it would be friendlier for user, but I can make it only uppercase. Is it better?

scripts/install.sh
utku-caglayan: Yes, thanks for catching that. Will fix it as discussed

PR #32, opened 2021-11-23
Introduce interactive clc mode [API-689]

utku-caglayan:


commands/root.go
```diff
@@ -16,55 +16,131 @@
...
 			return cmd.Help()
 		},
 	}
 )
 
+func addressAndClusterNamePrefix() (prefix string, useLive bool) {
+	if config := internal.Configuration; config != nil {
+		return fmt.Sprintf("hzc %s@%s> ", config.Cluster.Network.Addresses[0], config.Cluster.Name), true
+	}
+	return "hzc address@clusterName> ", true
```
yuce, line 55: Is it probable that the configuration is `nil`? I think when this function is called, we should have already connected to the cluster, shouldn't we?

utku-caglayan: Thank you, fixed as suggested

commands/root.go
```diff
@@ -16,55 +16,131 @@
...
+var advancedPrompt = &cobraprompt.CobraPrompt{
+	RootCmd:                  RootCmd,
+	PersistFlagValues:        true,
+	ShowHelpCommandAndFlags:  true,
+	ShowHiddenFlags:          true,
+	SuggestFlagsWithoutDash:  true,
+	DisableCompletionCommand: true,
+	AddDefaultExitCommand:    true,
+	GoPromptOptions: []prompt.Option{
+		prompt.OptionTitle("hazelcast commandline client"),
```
yuce, line 64: Is this title displayed as the title of the terminal window? How about: "Hazelcast Commandline Client" ?

commands/root.go
```diff
@@ -16,55 +16,131 @@
...
+	}
+	var flagsToExclude []string
+	RootCmd.PersistentFlags().VisitAll(func(flag *pflag.Flag) {
+		flagsToExclude = append(flagsToExclude, flag.Name)
+	})
+	flagsToExclude = append(flagsToExclude, "help")
+	advancedPrompt.FlagsToExclude = flagsToExclude
+	RootCmd.Example = "> map put -k key -m myMap -v someValue\n" +
+		"> map get -k key -m myMap\n" +
+		"> cluster version"
```
yuce, line 122: Instead of concatenating strings and adding line breaks, backticks (````) may be used.

commands/root.go
```diff
@@ -16,55 +16,131 @@
...
+}
+
 func decorateRootCommand(cmd *cobra.Command) {
-	cmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", fmt.Sprintf("config file (default is %s)", internal.DefautConfigPath()))
-	cmd.PersistentFlags().StringVarP(&addresses, "address", "a", "", "addresses of the instances in the cluster.")
-	cmd.PersistentFlags().StringVarP(&cluster, "cluster-name", "", "", "name of the cluster that contains the instances.")
-	cmd.PersistentFlags().StringVar(&token, "cloud-token", "", "your Hazelcast Cloud token.")
+	cmd.PersistentFlags().StringVarP(&internal.CfgFile, "config", "c", internal.DefautConfigPath(), fmt.Sprintf("config file (default is %s)", internal.DefautConfigPath()))
+	cmd.PersistentFlags().StringVarP(&internal.Address, "address", "a", "", "addresses of the instances in the cluster.")
+	cmd.PersistentFlags().StringVarP(&internal.Cluster, "cluster-name", "n", "", "name of the cluster that contains the instances.")
```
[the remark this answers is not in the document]
utku-caglayan: Yes correct, I removed the shorthand for this.

internal/cobraprompt/prompt.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	if co.PersistFlagValues {
+		co.RootCmd.PersistentFlags().BoolP(PersistFlagValuesFlag, "",
+			false, "Persist last given value for flags")
+	}
+}
+
+func findSuggestions(co *CobraPrompt, d *prompt.Document) []prompt.Suggest {
+	command := co.RootCmd
+	args := strings.Fields(d.CurrentLine())
```
yuce, line 112: Will this cause a problem with args which have spaces?

internal/cobraprompt/prompt.go
```diff
@@ -0,0 +1,164 @@
...
+	if found, _, err := command.Find(args); err == nil {
+		command = found
+	}
+	var suggestions []prompt.Suggest
+	persistFlagValues, _ := command.Flags().GetBool(PersistFlagValuesFlag)
```
yuce, line 117: The error is ignored here. Is that intended?

commands/root.go
utku-caglayan: Yes it is, sure that makes sense. I would also prefer a shorter name since there are no guarantees that we will have sufficient space to display the whole title. What do you think of "Hazelcast Cli"?

commands/root.go
utku-caglayan: Great catch, thanks

internal/cobraprompt/prompt.go
utku-caglayan: Great catch, thanks. Fixed it with an external lib

commands/root.go
yuce: How about "Hazelcast Client"?

internal/cobraprompt/prompt.go
utku-caglayan: I tried to add a workaround, because initiation of commands etc.. are a mess because of the global state. Changes are on [28bd268](https://github.com/hazelcast/hazelcast-commandline-client/pull/32/commits/28bd268ea81423605d78b702bc754105544cbd3c). I will keep this in mind for a refactor session if that is okay with you too.

yuce (approved):
Thanks, looks good!

PR #38, opened 2022-01-04
Compact Code Generator

sancar:


sancar:
We have decided to make it a separate library. Closing the pr for now.

PR #39, opened 2022-01-14
SQL support [API-1084]

utku-caglayan:
Initial effort to support sql via 'hzc sql query "some query"' command. It works both interactive and non-interactive way.


utku-caglayan:
!! This probably will be closed since https://github.com/hazelcast/hazelcast-commandline-client/pull/40
builds on top of this !!

yuce:
@utku-caglayan Can we close this one?

PR #40, opened 2022-01-23
Interactive SQL [API-1084]

utku-caglayan:
- implement sql browser with system-wide pager command
- add context cancellation support


internal/cobraprompt/context.go
```diff
@@ -0,0 +1,24 @@
+package cobraprompt
+
+import (
+	"context"
+	"time"
+)
+
+// CobraMutableCtx is a workaround for https://github.com/spf13/cobra/issues/1109
+type CobraMutableCtx struct {
```
yuce, line 9: It's not a good idea to implement the Context interface. Can't we use `context.Context` in the standard lib?

main.go
```diff
@@ -16,14 +16,17 @@
...
+
 	"github.com/hazelcast/hazelcast-commandline-client/commands"
 )
 
 func main() {
+	ctx := context.Background()
 	isInteractive := commands.IsInteractiveCall()
 	if isInteractive {
-		commands.ExecuteInteractive()
+		commands.ExecuteInteractive(ctx)
```
yuce, line 28: You can `return` after this line to be able to remove the `else` in the line below.

yuce (changes requested):
A few comments, reviewing more...

internal/cobraprompt/prompt.go
```diff
@@ -75,21 +76,38 @@ func handleExit() {
 
 // Run will automatically generate suggestions for all cobra commands and flags defined by RootCmd and execute the selected commands.
 // Run will also reset all given flags by default, see PersistFlagValues
-func (co CobraPrompt) Run(ctx context.Context) {
+func (co CobraPrompt) Run(cntx context.Context) {
 	defer handleExit()
+	ctx, cancelFunc := context.WithCancel(cntx)
```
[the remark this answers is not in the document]
utku-caglayan: Thanks, fixed at https://github.com/hazelcast/hazelcast-commandline-client/pull/40/commits/e5b6f7a183de2c648671a2ffa62fad00b79baf62

internal/cobraprompt/prompt.go
```diff
@@ -75,21 +76,38 @@ func handleExit() {
...
 		panic("RootCmd is not set. Please set RootCmd")
 	}
 	co.prepare()
+	var (
+		cobraCtx  CobraMutableCtx
+		cmdCancel context.CancelFunc
+	)
 	p := prompt.New(
 		func(in string) {
+			cobraCtx.Internal, cmdCancel = context.WithCancel(ctx)
```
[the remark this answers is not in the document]
utku-caglayan: Thanks, fixed at https://github.com/hazelcast/hazelcast-commandline-client/pull/40/commits/e5b6f7a183de2c648671a2ffa62fad00b79baf62

main.go
utku-caglayan: That's better thanks, fixed at https://github.com/hazelcast/hazelcast-commandline-client/pull/40/commits/e5b6f7a183de2c648671a2ffa62fad00b79baf62

internal/sql/db.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func makeDSN(config *hazelcast.Config) string {
+	ll := logger.OffLevel
+	q := url.Values{}
+	q.Add("cluster.name", config.Cluster.Name)
+	q.Add("unisocket", strconv.FormatBool(config.Cluster.Unisocket))
+	q.Add("log", string(ll))
+	return fmt.Sprintf("hz://%s?%s", config.Cluster.Network.Addresses[0], q.Encode())
+}
```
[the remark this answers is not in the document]
utku-caglayan: Definitely! Fixed at https://github.com/hazelcast/hazelcast-commandline-client/pull/40/commits/10929023b68a8f291c0288b532f8f17709047666 with some light refactoring 🤭

internal/table/console_test.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	exampleCells := []interface{}{"someValue", 5.0, time.Date(1994, 8, 30, 0, 0, 0, 0, time.UTC)}
+	exampleLine := "|someValue      |5              |1994-08-30 0...|"
+	buffer := bytes.NewBuffer(nil)
+	reader, writer := bufio.NewReader(buffer), buffer
+	w := NewTableWriter(writer)
+	steps := []struct {
+		desc               string
+		cells              []interface{}
+		changeTerminalSize func()
```
emreyigit, line 24: Is `changeTerminalSize` some kind future investment? I couldn't find where and how it is used.

internal/table/console.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+/*
+WriteHeader outputs header for the table with the form:
++--------------------------------------+
+| vegetables | fruit      | rank       |
++--------------------------------------+
+*/
+func (t *TabularWriter) WriteHeader(cells ...interface{}) error {
```
emreyigit, line 69: Just wondering, will you plan to use same function to produce output files or, what if I direct `>` the output to a file? In this case, how trimmed columns (`longlongvegetables` -> `longlongvege...`) will be managed.

internal/table/console_test.go
utku-caglayan: Yes, I know it is not used right know but I wanted to include it just as a reminder to write tests that manipulate terminal size in the future. If you feel uncomfortable with that approach, I can gladly revert it 😊

internal/table/console.go
utku-caglayan: Great catch! I imagine it is not a use case for interactive browsing. This would be very useful for non-interactive scripting purposes. For that we are considering adding a "--csv" flag such as:
```hzc sql query --csv "SQLQUERY" ```
which will output every field and column without applying any wrapping.

And for now, I do not think we should support it on interactive mode since we are trying to separate those use cases.

emreyigit: I got it, thanx. Than, there is no problem. While talking about flags, can we able to set flags also from environment variable?

utku-caglayan: Currently we do not have support for that. I think it is low effort to do, but I am not sure how common it is on linux environment to manipulate flags via environment variable.
If you come across something like that, I think it would be great to consider it 🙏

internal/cobraprompt/context.go
yuce: We'll use it as is for now.

yuce (approved):
Looks good!

emreyigit (approved):

PR #41, opened 2022-01-27
Demo fixes [API-1117]

utku-caglayan:
- Add interactive shell example to “example”
- Hide persistent flags such as —cluster-token in help text in interactive mode
- Fix output for “map get —value-type json”
- Exclude already used flags from suggestions
- Partially fix suggestion when input is not valid

### DIFF
--- a/commands/types/map/mapget.go
+++ b/commands/types/map/mapget.go
@@ -57,7 +57,7 @@ var mapGetCmd = &cobra.Command{
 		if value != nil {
 			switch v := value.(type) {
 			case serialization.JSON:
-				if err := quick.Highlight(os.Stdout, v.String(),
+				if err := quick.Highlight(os.Stdout, fmt.Sprintln(v.String()),
 					"json", "terminal", "tango"); err != nil {
 					fmt.Println(v.String())
 				}
--- a/internal/cobraprompt/prompt.go
+++ b/internal/cobraprompt/prompt.go
@@ -12,6 +12,8 @@ import (
 	"github.com/google/shlex"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+
+	"github.com/hazelcast/hazelcast-commandline-client/internal"
 )
 
 // DynamicSuggestionsAnnotation for dynamic suggestions.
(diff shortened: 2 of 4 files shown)


internal/cobraprompt/prompt.go
```diff
@@ -151,54 +153,67 @@ func RegisterPersistFlag(co *cobra.Command) {
...
 			return
 		}
 		if stringInSlice(co.FlagsToExclude, flag.Name) {
 			return
 		}
 		flagUsage := "--" + flag.Name
-		if strings.HasPrefix(d.GetWordBeforeCursor(), "--") {
+		// Check if flag is already used in the command
+		if (flag.Shorthand != "" && internal.IsOneOf(words, "-"+flag.Shorthand, true)) ||
+			internal.IsOneOf(words, flagUsage, true) {
```
yuce, line 202: How about introducing the `internal/check` package for `IsOneOf` just like we do for the Go client?

internal/util.go
```diff
@@ -0,0 +1,17 @@
+package internal
+
+import "strings"
+
+func IsOneOf(list []string, s string, caseSensitive bool) bool {
```
yuce, line 5: How about renaming this function to `ContainsString` and removing the `caseSensitive` parameter?

We can introduce a new `ContainsStringCaseInsensitive` function which checks compares with case sensitivity.

internal/cobraprompt/prompt.go
utku-caglayan: Sounds good. Fixed at https://github.com/hazelcast/hazelcast-commandline-client/pull/41/commits/0bdfd09792e7af0022e3b8ab244bec7b2813bd57

internal/util.go
utku-caglayan: Fixed at https://github.com/hazelcast/hazelcast-commandline-client/pull/41/commits/0bdfd09792e7af0022e3b8ab244bec7b2813bd57

Will introduce case insensitive version when it is needed 👍

yuce (approved):

commands/types/map/mapget.go
```diff
@@ -57,7 +57,7 @@ var mapGetCmd = &cobra.Command{
 		if value != nil {
 			switch v := value.(type) {
 			case serialization.JSON:
-				if err := quick.Highlight(os.Stdout, v.String(),
+				if err := quick.Highlight(os.Stdout, fmt.Sprintln(v.String()),
```
nevzatseferoglu, line 60: I don't understand why you wrap the string around the `Sprintln`. The new [PR](https://github.com/utku-caglayan/hazelcast-commandline-client/blob/improvements-1/commands/types/map/mapget.go#L59) is handling other way around.

nevzatseferoglu (commented):
Looks good, left a comment.

nevzatseferoglu (approved):
Looks good.

yuce:
@utku-caglayan Is there anything keeping this PR get merged?

PR #42, opened 2022-01-27
Vendor [API-1117]

utku-caglayan:
- vendor dependencies
- fix history index out of range bug
- select suggestion on enter
- append space after suggestion selection


srknzl:
Why do we add vendor folders and what is it? Is it like committing node_modules?

utku-caglayan:
> Why do we add vendor folders and what is it? Is it like committing node_modules?

Yes exactly, we did some changes the original dependency under vendor directory.

utku-caglayan:
This is closed to be continued with a different approach on #58. Thanks for the reviews!

PR #43, opened 2022-01-28
Add SSL Config

yuce:


internal/connect.go
```diff
@@ -148,7 +147,7 @@ func ConnectToCluster(ctx context.Context, clientConfig *hazelcast.Config) (cli
 	defer cancel()
 	configCopy := clientConfig.Clone()
 	// prevent internal event loop to print error logs
-	configCopy.Logger.Level = logger.OffLevel
+	//configCopy.Logger.Level = logger.OffLevel
```
utku-caglayan, line 150: I think we can move this to newly added "DefaultConfig". Wdyt?

utku-caglayan (changes requested):

internal/connect.go
yuce: I had to comment out that line in order to be able to see the log. Probably `DefaultConfig` doesn't work as intended?

utku-caglayan: You are correct! It should suffice if we can change the default log level here:
https://github.com/yuce/hazelcast-commandline-client/blob/ssl-config/internal/config.go#L67
`config.Logger.Level = logger.OffLevel`

This would initiate a default config on user system with client level logging disabled. Also we need to remove the line you correctly commented out above.

Sorry for this, I should have handled this better

internal/config.go
```diff
@@ -143,3 +161,21 @@ func DefautConfigPath() string {
 	}
 	return filepath.Join(homeDirectoryPath, ".local/share/hz-cli", defaultConfigFilename)
 }
+
+func UpdateConfigWithSSL(config *hazelcast.Config, sslc *SSLConfig) error {
+	if sslc.ServerName != "" && sslc.InsecureSkipVerify {
+		return fmt.Errorf("SSL.ServerName and SSL.InsecureSkipVerify are mutually exclusive")
```
utku-caglayan, line 167:
```suggestion
		return fmt.Errorf("SSL.ServerName and SSL.InsecureSkipVerify are mutually exclusive\n")
```

yuce: I think we shouldn't put formatting code in the error messages. Is `\n` necessary for other error messages as well?

utku-caglayan: My bad, I commented on wrong place

internal/config.go
```diff
@@ -113,27 +130,28 @@ func MakeConfig() (*hazelcast.Config, error) {
 		return nil, fmt.Errorf("error reading Configuration at %s: %w", confPath, err)
 	}
 	if Token != "" {
-		config.Cluster.Cloud.Token = strings.TrimSpace(Token)
-		config.Cluster.Cloud.Enabled = true
+		config.Hazelcast.Cluster.Cloud.Token = strings.TrimSpace(Token)
+		config.Hazelcast.Cluster.Cloud.Enabled = true
+	}
+	if err := UpdateConfigWithSSL(&config.Hazelcast, &config.SSL); err != nil {
+		return nil, err
```
utku-caglayan, line 137: Errors on MakeConfig are logged in the function since we do not have contextual information on "command" level about the errors.
To be in sync with rest of the flow, can you log the error here? Non-interactive mode does not print any error when this line is triggered

README.md
```diff
@@ -135,6 +135,52 @@ The cluster creation and retrieving connection info can be done directly in comm
   hzc --cluster-name <CLUSTER NAME> --cloud-token <DISCOVERY TOKEN>
   ```
 
+## SSL Configuration
+
+You can use the following configuration file to enable SSL support:
+```
+ssl:
+    servername: "HOSTNAME-FOR-SERVER"
+    # or: insecureskipverify: true
```
srknzl, line 144: maybe we can delete `or:` part from this comment. also do we have a place where these config options are documented. we can point to that documentation in CLC docs.

README.md
```diff
@@ -135,6 +135,52 @@ The cluster creation and retrieving connection info can be done directly in comm
   hzc --cluster-name <CLUSTER NAME> --cloud-token <DISCOVERY TOKEN>
   ```
 
+## SSL Configuration
+
+You can use the following configuration file to enable SSL support:
+```
+ssl:
```
srknzl, line 142: We should try all these scenarios to see if they work

README.md
```diff
@@ -135,6 +135,52 @@ The cluster creation and retrieving connection info can be done directly in comm
...
+
+Cloud SSL configuration:
+```
+ssl:
+    servername: "hazelcast.cloud"
+    capath: "/tmp/ca.pem"
+    certpath: "/tmp/cert.pem"
+    keypath: "/tmp/key.pem"
+    keypassword: "PASSWORD FOR THE KEY"
+hazelcast: #
```
srknzl, line 177: we can delete this
```suggestion
hazelcast:
```

internal/config.go
```diff
@@ -36,6 +37,22 @@ const (
...
+	InsecureSkipVerify bool
+	CAPath             string
+	CertPath           string
+	KeyPath            string
+	KeyPassword        string
+}
+
+type Config struct {
+	Hazelcast hazelcast.Config
+	SSL       SSLConfig
```
srknzl, line 51: Why these are next to each other and SSL is not child of Hazelcast?

internal/config.go
```diff
@@ -143,3 +161,41 @@ func DefautConfigPath() string {
...
+		return fmt.Errorf("SSL.ServerName and SSL.InsecureSkipVerify are mutually exclusive")
+	}
+	var tlsc *tls.Config
+	if sslc.ServerName != "" {
+		tlsc = &tls.Config{ServerName: sslc.ServerName}
+	} else if sslc.InsecureSkipVerify {
+		tlsc = &tls.Config{InsecureSkipVerify: true}
+	}
+	csslc := &config.Cluster.Network.SSL
+	if tlsc != nil {
```
srknzl, line 179: If a user sets `config.Cluster.Network.SSL` to false and `SSL.ServerName` to something so that `csslc.Enabled` is set to true? Is this documented? They may want to temporariliy disable SSL without deleting SSL part in yaml.

internal/config.go
```diff
@@ -143,3 +161,41 @@ func DefautConfigPath() string {
 	}
 	return filepath.Join(homeDirectoryPath, ".local/share/hz-cli", defaultConfigFilename)
 }
+
+func UpdateConfigWithSSL(config *hazelcast.Config, sslc *SSLConfig) error {
+	if sslc.ServerName != "" && sslc.InsecureSkipVerify {
+		return fmt.Errorf("SSL.ServerName and SSL.InsecureSkipVerify are mutually exclusive")
```
srknzl, line 170: I believe we should document that these are mut exclusive.

internal/config.go
yuce: Updated at: https://github.com/hazelcast/hazelcast-commandline-client/pull/43/commits/56e501f45372988d99cf1bf7b0efa796cf365255

README.md
yuce: The idea was making the user aware of `insecureskipverify` option, which can be used when `servername` cannot be used. README is the only documentation for CLC. We should improve that.

internal/config.go
yuce: The `SSL` configuration of CLC overrides Go client SSL config.

internal/config.go
yuce: Go client doesn't support most of the SSL settings declaratively, so I had to add an extra config section for CLC.

README.md
yuce: Updated at: https://github.com/hazelcast/hazelcast-commandline-client/pull/43/commits/7e85138d4528191eb23ce12620e392464ac1a663

README.md
yuce: We tried them manually and they seem to work.

internal/config.go
srknzl: It is ok to do that but we should improve documentation. Will you do later?

srknzl (approved):

internal/config.go
yuce: You are right, we will add better documentation during refactoring.

utku-caglayan (approved):

PR #55, opened 2022-02-10
[EASY] change ghExtractTag for newer release tags

utku-caglayan:
Currently installation script cannot parse release tag correctly. To reproduce to error and fix, you can try
```
x="https://github.com/hazelcast/hazelcast-commandline-client/releases/tag/v0.1.0-preview.1"
echo ${x##*v}
echo ${x##*tag/v}
```
on bash terminal

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -5,7 +5,7 @@ HZCLI_HOME="$HOME/.local/share/hz-cli"
 
 ghExtractTag() {
   tagUrl=$(curl "https://github.com/$1/releases/latest" -s -L -I -o /dev/null -w '%{url_effective}')
-  printf "%s\n" "${tagUrl##*v}"
+  printf "%s\n" "${tagUrl##*tag/v}"
 }
 
 bin_id=""


yuce (approved):

scripts/install.sh
```diff
@@ -5,7 +5,7 @@ HZCLI_HOME="$HOME/.local/share/hz-cli"
 
 ghExtractTag() {
   tagUrl=$(curl "https://github.com/$1/releases/latest" -s -L -I -o /dev/null -w '%{url_effective}')
-  printf "%s\n" "${tagUrl##*v}"
```
nevzatseferoglu, line 8: `x="https://github.com/hazelcast/hazelcast-commandline-client/releases/tag/vv0.1.0-preview.1"`

But it will also parse as if it corrects, does it create a worry ?

nevzatseferoglu: I guess we are retrieving URL from the `curl`, That case will not be happened.

utku-caglayan: Yes this assumes we will use semver for all tags.

PR #64, opened 2022-04-28
Fix Ctrl-left behavior [API-1208]

utku-caglayan:
We observed that ctrl - left did not work as expected. This work tries to fix that. I confirmed that it looks good on:
- zsh with some tweaks to send correct char sequences for ctrl+arrow keys. By default it does not use readline() under the hood. This fix is necessary to capture those.

Dear reviewers, if you have an environment other than this, please also try and see if it is fixed. Thanks!

### DIFF
--- a/internal/go-prompt/document.go
+++ b/internal/go-prompt/document.go
@@ -2,6 +2,7 @@ package prompt
 
 import (
 	"strings"
+	"unicode"
 	"unicode/utf8"
 
 	"github.com/mattn/go-runewidth"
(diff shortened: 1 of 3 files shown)


internal/go-prompt/document.go
```diff
@@ -198,6 +199,23 @@ func (d *Document) FindEndOfCurrentWord() int {
 	return len(x)
 }
 
+func (d *Document) FindPreviousWordStart() int {
```
yuce, line 202: Could you add a test for this method?

yuce (changes requested):

internal/go-prompt/document.go
utku-caglayan: Thanks, added at https://github.com/hazelcast/hazelcast-commandline-client/pull/64/commits/60b613b79be4e727690ba65811c2589193df16ce

internal/go-prompt/document_test.go
```diff
@@ -0,0 +1,78 @@
...
+			},
+			charsToLeft: 4,
+		},
+		{
+			doc: Document{
+				Text:           "abc xw",
+				cursorPosition: 5,
+			},
+			charsToLeft: 1,
+		},
```
nevzatseferoglu, line 70:
```
doc: Document{
Text:       "    ",
    cursorPosition: 1,
},
charsToLeft: 0,
```
I'm not sure that it is the correct behavior or not but the test above returns one as expected even there is no word before the given cursor position. Shouldn't have it returned zero ?

nevzatseferoglu: Since it finds the previous word starting position through current index of the curser according to whitespaces. It is clear that when we type sequential spaces then press `ctrl + left arrow`, it should return back and jump over all contiguous spaces to the beginning the given literal. I'm resolving.

yuce (approved):
Tried it on Ubuntu 20.04 + Bash. Looks good!

PR #73, opened 2022-06-01
Phone home support [API-1355]

utku-caglayan:
- Changed go client version to leverage from the latest developments with versioning via build flags
- Modified makefile and goreleaser to build with client type name and version

### DIFF
--- a/internal/connect.go
+++ b/internal/connect.go
@@ -34,6 +34,7 @@ import (
 
 var InvalidStateErr = errors.New("invalid new state")
 
+//todo add protection for concurrent access
 var (
 	client    *hazelcast.Client
 	sqlDriver *sql.DB
--- a/internal/connect.go
+++ b/internal/connect.go
@@ -17,6 +17,7 @@ package internal
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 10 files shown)


internal/connect.go
```diff
@@ -32,7 +34,10 @@ import (
 
 var InvalidStateErr = errors.New("invalid new state")
 
-var client *hazelcast.Client
+var (
+	client    *hazelcast.Client
+	sqlDriver *sql.DB
```
yuce, line 39: Should these global variables protected against concurrent access?

yuce (commented):
Looks good, just a question...

internal/connect.go
utku-caglayan: No they are not, but because of the nature of CLI, only 1 command tries to access them in any point in time. I can introduce them if you'd like, but didn't think it was necessary so far

yuce: OK. Could you add a TODO to protect those? In the later releases we may introduce things like background jobs.

srknzl (dismissed):

srknzl (approved):

yuce (approved):

yuce (approved):
Looks good!

srknzl (approved):

PR #76, opened 2022-06-06
Refactor an example for put-all

utku-caglayan:


yuce (dismissed):

types/mapcmd/put-all.go
```diff
@@ -38,8 +38,8 @@ const MapPutAllExample = `  # Put key, value pairs to map.
   # Put key, value pairs to map in another order.
   hzc map put-all -n mapname -k k1 -k k2 -v v1 -v v2
   
-  # Put key, value pairs to map but one of the value type is json file.
-  hzc map put-all -n mapname -k k1 -f valueFile.json -t json -k k2 -v v2
+  # Put key, value pairs to map with value types as JSON
+  hzc map put-all -n mapname -k k1 -f valueFile.json -k k2 -v '{"field":"tmp"}' -t json
```
srknzl, line 42: What is the type of first key? How is it determined? Don't we need -t parameter for that as well?

srknzl (approved):

srknzl (approved):

types/mapcmd/put-all.go
```diff
@@ -38,8 +38,11 @@ const MapPutAllExample = `  # Put key, value pairs to map.
   # Put key, value pairs to map in another order.
   hzc map put-all -n mapname -k k1 -k k2 -v v1 -v v2
   
-  # Put key, value pairs to map but one of the value type is json file.
-  hzc map put-all -n mapname -k k1 -f valueFile.json -t json -k k2 -v v2
+  # Put key, value pairs to map with value types as JSON (default type string)
+  hzc map put-all -n mapname -k k1 -f valueFile.json -k k2 -v '{"field":"tmp"}' -t json
```
yuce, line 42: How about having a single example for put-all and having more examples in the documentation?

PR #78, opened 2022-06-07
Simplify examples

utku-caglayan:
Some examples may be lengthy for command line interface (small window size)

### DIFF
--- a/types/mapcmd/get.go
+++ b/types/mapcmd/get.go
@@ -25,8 +25,7 @@ import (
 )
 
 const MapGetExample = `  # Get value of the given key from the map.
-  hzc map get --key hello --name myMap   # default key-type is string
-  hzc map get --key-type int16 --key 2012 --name yearbook`
+  hzc map get --key-type int16 --key 2012 --name myMap   # default key-type is string`
 
 func NewGet(config *hazelcast.Config) *cobra.Command {
 	var mapName, mapKey, mapKeyType string
--- a/types/mapcmd/get.go
+++ b/types/mapcmd/get.go
@@ -25,8 +25,7 @@ import (
 )
 
 const MapGetExample = `  # Get value of the given key from the map.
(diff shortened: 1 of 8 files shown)


types/mapcmd/get.go
```diff
@@ -25,8 +25,7 @@ import (
 )
 
 const MapGetExample = `  # Get value of the given key from the map.
-  hzc map get -n mapname -k k1
-  hzc map get --key hello --name myMap
+  hzc map get --key hello --name myMap   # default key-type is string
```
yuce, line 28: Do you think we can add the `key-type` here and get rid of the example below?

types/mapcmd/put.go
```diff
@@ -26,35 +26,11 @@ import (
 	"github.com/hazelcast/hazelcast-commandline-client/internal"
 )
 
-const MapPutExample = `  # Put key, value pair to map.
-  hzc put -n mapname -k k1 -v v1
+const MapPutExample = `  # Put key, value pair to map. ttl/max-idle can be (ns,us,ms,s,m,h)
```
yuce, line 29: How about "The unit for ttl/max-idle is one of (ns,us,ms,s,m,h)"

types/mapcmd/put.go
```diff
@@ -26,35 +26,11 @@ import (
...
+  map put --key-type string --key hello --value-type float32 --value 19.94 --name myMap --ttl 3ms --max-idle 4ms
```
yuce: How about using `s` and `m` for TTL and max idle? I think TTL < 1 is not legal.

yuce (commented):
Some ideas...

types/mapcmd/get.go
utku-caglayan: How about just keeping the example below?

alisengul53 (approved):

PR #80, opened 2022-06-08
[Trivial] Update cloud ssl configuration example

utku-caglayan:


README.md
```diff
@@ -205,6 +205,7 @@ hazelcast:
 Cloud SSL configuration:
 ```
 ssl:
+    servername: "hazelcast.cloud"
```
yuce, line 208: How about setting this by default if the cloud is enabled, so the user doesn't have to configure it?

config/config.go
```diff
@@ -92,6 +92,9 @@ func mergeFlagsWithConfig(flags *GlobalFlagValues, config *Config) error {
 	if flags.Token != "" {
 		config.Hazelcast.Cluster.Cloud.Token = strings.TrimSpace(flags.Token)
 		config.Hazelcast.Cluster.Cloud.Enabled = true
+		if config.SSL.ServerName == "" && !config.SSL.InsecureSkipVerify {
```
yuce, line 95: I think it's OK to not check for `InsecureSkipVerify`. Setting server name shouldn't affect that one.

config/config.go
```diff
@@ -154,6 +151,11 @@ func DefaultConfigPath() string {
 }
 
 func updateConfigWithSSL(config *hazelcast.Config, sslc *SSLConfig) error {
+	var emptySSLConfig SSLConfig
+	if *sslc == emptySSLConfig {
```
yuce, line 155: How about: `*sslc == SSLConfig{}`

yuce (approved):
Looks good!

alisengul53 (approved):

PR #81, opened 2022-06-09
Integration Test

yuce:
This PR:
- Fixes map commands, so they output to stdout.
- Adds an expectation test.

The test clears `test-map`, adds 1000 entries using separate `map put` commands and retrives them all with `map get-all`. If the number of retrieved entries is not 1000 or a non-expected value received for a key, the test fails.


utku-caglayan (approved):
Looks good thanks

alisengul53 (approved):

Issue #99, opened 2022-07-26
Can't exit from SQL screen  [API-1476]


utku-caglayan:
Can't exit from SQL browser when a query is running and the member shuts down in the middle of it

PR #100, opened 2022-07-26
Fix connection bug to non-existing members on SQL screen [API-1466]

### LINKED ISSUE #85: Cannot terminate the process
1. Start a local member on localhost:5701
2. Execute `hzc sql -a 127.0.0.1:5702` to try to connect to a non-existent member.

The logs are flooded with error messages and Ctrl + C does not work to terminate the process. I had to force close the Terminal window.

utku-caglayan:
fixes #85 

There is also a bug #99  this does not address it

### DIFF
--- a/internal/connect.go
+++ b/internal/connect.go
@@ -161,11 +161,11 @@ func ConnectToCluster(ctx context.Context, clientConfig *hazelcast.Config) (cli 
 	return
 }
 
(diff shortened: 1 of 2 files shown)


yuce (approved):

mehmettokgoz (approved):
Thanks.

PR #126, opened 2022-08-13
Post v5.2.0-beta2 Fixes

yuce:
CLC v5.2.0-beta2 puts the `network` configuration in the same level as `cluster` in the default configuration, which is not correct. Fixed that and:

- Refactored default config.
- Make sure the default config is correct.
- Removed less important settings from the default configuration.
- Removed Windows build from goreleaser.
- Removed unnecessary files from gorelaser build.
- Raised min Go to 1.18.
- Added version suffix to Windows installer's name.
- Updated README to reflect the latest Windows related improvements.


config/config.go
```diff
@@ -71,55 +70,48 @@ type GlobalFlagValues struct {
 	NoColor          bool
 }
 
-func DefaultConfig() *Config {
+func DefaultConfig() Config {
```
mehmettokgoz, line 73: Why did we convert return type to object instead of pointer?

config/config.go
```diff
@@ -71,55 +70,48 @@ type GlobalFlagValues struct {
...
 styling:
-  theme: "default" # default, no-color, solarized
-  colorpalette:
-    # uncomment to override theme color. closest supported color will be used
-    #headerbackground: "#ff12aa"
-    #border: "#ff12aa"
-    #resulttext: "#ff12aa"
-    #headerforeground: "#000000"
-    #highlight: "#ff12aa"
-    #footerforeground: "#ff12aa"
```
mehmettokgoz, line 122: I'm assuming we are removing the color options intentionally.

config/config_test.go
```diff
@@ -155,12 +149,56 @@ func TestMergeFlagsWithConfig(t *testing.T) {
...
+	assert.Equal(t, true, cc.Unisocket)
+	assert.Equal(t, cluster.NetworkConfig{Addresses: []string{"localhost:5701"}}, cc.Network)
+	assert.Equal(t, cluster.CloudConfig{}, cc.Cloud)
+	assert.Equal(t, cluster.SecurityConfig{}, cc.Security)
+	assert.Equal(t, cluster.DiscoveryConfig{}, cc.Discovery)
+	assert.Equal(t, logger.ErrorLevel, cfg.Hazelcast.Logger.Level)
+	assert.Equal(t, SSLConfig{}, cfg.SSL)
+	assert.Equal(t, false, cfg.NoAutocompletion)
+	assert.Equal(t, "default", cfg.Styling.Theme)
+
```
mehmettokgoz, line 187: Unnecessary blank line I guess.

config/config.go
yuce: In most cases it's better to return a non-pointer value. In this specific case, returning the address of `Config` causes it to escape to heap, but returning it doesn't (we don't care about performance in many code paths in CLC, so that's not very important).

config/config.go
yuce: The color options are removed only from the default configuration to make the default config contain the most important options. The user can still add and use them.

config/config_test.go
yuce: Removed at: https://github.com/hazelcast/hazelcast-commandline-client/pull/126/commits/9cbbce81c69bbc73b5e97d665adc3948277d6647

mehmettokgoz (approved):
Looks good to me, thanks!

Issue #96, opened 2022-07-19
Can't disable/switch from Interactive mode [API-1477]


angelasimms:
Hi,

I'm running Ubuntu on Windows. 

I can enable interactive mode from the bash: `hzc #` but I can't disable it. I've tried CTRL + C, exit, complete shutdown, renaming my bash profile... I've also tried switching to SQL mode.

yuce:
Tracking this at: https://hazelcast.atlassian.net/browse/API-1455

utku-caglayan:
@angelasimms thanks for reporting. We are not exaclty sure what is the scenario but hopefully #132 fixed the issue. Please feel free to re-open if you experience it again.

PR #132, opened 2022-09-06
Fix CLC can't exit from SQL queries when cluster is down [API-1477] [API-1476]

### LINKED ISSUE #99: Can't exit from SQL screen  [API-1476]

utku-caglayan:
For the following scenario, CLC can not exit from the interactive mode and the SQL Browser mode.

1. Run a streaming query such as "select * from table(generate_table(1))"
2. Kill test cluster
3. Try to exit

This work addresses that by replacing the Go Client SQL driver with SQL service API with some additional changes.

Fixes #99


mehmettokgoz (approved):
Look good to me, thanks!

utku-caglayan:
Fixes #96

PR #133, opened 2022-09-07
Add list-objects command [API-1517]

utku-caglayan:
Adds list-objects command. There are few issues to be discussed. 
- Flake ID Generator's type is received as "PN Counter"
- Topic's type is received as "Replicated Map"

### DIFF
--- a/list_objects/list.go
+++ b/list_objects/list.go
@@ -1,4 +1,4 @@
-package listcmd
+package list_objects
 
 import (
 	"context"
--- a/list_objects/list_test.go
+++ b/list_objects/list_test.go
@@ -1,4 +1,4 @@
-package listcmd
+package list_objects
 
 import (
 	"bytes"
--- a/listcmd/list.go
+++ b/listcmd/list.go
@@ -1,0 +1,123 @@
+package listcmd
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/hazelcast/hazelcast-go-client"
+	"github.com/spf13/cobra"
+
+	hzcerrors "github.com/hazelcast/hazelcast-commandline-client/errors"
+	"github.com/hazelcast/hazelcast-commandline-client/internal"
+)
+
+const (
+	Map              = "map"
+	ReplicatedMap    = "replicatedMap"
+	MultiMap         = "multiMap"
+	Queue            = "queue"
+	Topic            = "topic"
+	List             = "list"
+	Set              = "set"
+	PNCounter        = "PNCounter"
+	FlakeIDGenerator = "flakeIdGenerator"
+	// unsupported types by go client
+	Cache                = "cache"
+	EventJournal         = "eventJournal"
+	Ringbuffer           = "ringBuffer"
+	FencedLock           = "fencedLock"
+	ISemaphore           = "semaphore"
+	IAtomicLong          = "atomicLong"
+	IAtomicReference     = "atomicReference"
+	ICountdownLatch      = "countdownLatch"
+	CardinalityEstimator = "cardinalityEstimator"
+)
+
+var dsObjTypes = []string{
+	Map,
+	ReplicatedMap,
+	MultiMap,
+	Queue,
+	Topic,
+	List,
+	Set,
+	PNCounter,
+	FlakeIDGenerator,
+	// unsupported types by go client
+	Cache,
+	EventJournal,
+	Ringbuffer,
+	FencedLock,
+	ISemaphore,
+	IAtomicLong,
+	IAtomicReference,
+	ICountdownLatch,
+	CardinalityEstimator,
+}
+
+var hzServiceToFilter = map[string]string{
+	hazelcast.ServiceNameMap:              Map,
+	hazelcast.ServiceNameReplicatedMap:    ReplicatedMap,
+	hazelcast.ServiceNameMultiMap:         MultiMap,
+	hazelcast.ServiceNameQueue:            Queue,
+	hazelcast.ServiceNameTopic:            Topic,
+	hazelcast.ServiceNameList:             List,
+	hazelcast.ServiceNameSet:              Set,
+	hazelcast.ServiceNamePNCounter:        PNCounter,
+	hazelcast.ServiceNameFlakeIDGenerator: FlakeIDGenerator,
+}
+
+func New(config *hazelcast.Config) *cobra.Command {
+	var objectType string
+	cmd := cobra.Command{
+		Use:   "list-objects [distributed object type]",
+		Short: "List distributed objects in the cluster",
+		Long:  `List type and name of the distributed objects present in the cluster`,
+		Example: `  list
+  list --type map
+  list --type fencedlock`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			c, err := internal.ConnectToCluster(ctx, config)
+			if err != nil {
+				return hzcerrors.NewLoggableError(err, "Can not connect to the cluster")
+			}
+			list, err := getObjects(ctx, c, objectType)
+			if err != nil {
+				return hzcerrors.NewLoggableError(err, "Can not get the distributed objects information")
+			}
+			sort.Strings(list)
+			l := strings.Join(list, "\n")
+			cmd.Println(l)
+			return nil
+		},
+	}
+	cmd.Flags().StringVarP(&objectType, "type", "t", "", fmt.Sprintf("type: %s", strings.Join(dsObjTypes, ",")))
+	cmd.RegisterFlagCompletionFunc("type", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return dsObjTypes, cobra.ShellCompDirectiveDefault
+	})
+	return &cmd
+}
+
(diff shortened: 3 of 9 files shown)


rootcmd/root.go
```diff
@@ -72,6 +73,7 @@ func subCommands(config *hazelcast.Config) []*cobra.Command {
 		mapcmd.New(config),
 		sqlcmd.New(config),
 		versioncmd.New(),
+		listcmd.New(config),
 	}
 	fds := []fakeDoor.FakeDoor{
 		{Name: "List", IssueNum: 48},
```
mehmettokgoz, line 79: Shouldn't we delete FakeDoor for List since CLC now supports it?

listcmd/list.go
```diff
@@ -0,0 +1,123 @@
+package listcmd
```
mehmettokgoz, line 1: Just a suggestion, `listcmd` looks like the List implementation at first sight, maybe we can change it to sth else. In future, we may want to implement the actual `listcmd` for the List data structure and it can create conflict.

utku-caglayan: Good idea! Renamed it https://github.com/hazelcast/hazelcast-commandline-client/pull/133/commits/9708369995df3bc8f2e73ac0358270d752507e50

list_objects/list_test.go
```diff
@@ -0,0 +1,116 @@
...
+		var b bytes.Buffer
+		cmd := New(&dc)
+		cmd.SetOut(&b)
+		ctx := context.Background()
+		cmd.SetArgs(tc.args)
+		_, err := cmd.ExecuteContextC(ctx)
+		require.NoError(t, err)
+		out := strings.Split(strings.TrimSpace(b.String()), "\n")
+		// at the of writing this, hazelcast creates an internal replicated map
+		// named "__sql.catalog". The line belows removes that
```
mehmettokgoz, line 103: I guess you mean `at the end of writing this`. Also, isn't this buffer holding the direct output of the `list-objects` command, why do we have to remove `__sql.catalog`? If it is there every time, I think we should remove it inside the command. I didn't get it.

list_objects/list_test.go
```diff
@@ -0,0 +1,116 @@
...
+	tcs := []struct {
+		name   string
+		args   []string
+		expect []string
+	}{
+		{
+			name: "list all",
+			args: []string{},
+			expect: []string{
+				// todo find out why flake id generators output as "PNCounter"
```
mehmettokgoz, line 70: I guess these are due to Go Client, not CLC.

mehmettokgoz:
Thanks, just a few questions.

list_objects/list_test.go
utku-caglayan: I'm not sure how to proceed about this. @yuce Do you think we should remove internal things "__someInternal.collection" in this command?

My concern is that users can name their structures starting with two underscores.

list_objects/list_test.go
utku-caglayan: Also cc @yuce

list_objects/list_test.go
yuce: I think it's OK to print all output for the moment. Maybe we can introduce a `--show-hidden` flag (or the opposite) in the future.

utku-caglayan: Thank you both, I'm resolving this for now.

list_objects/list.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var names []string
+	for _, t := range ts {
+		toFilter, ok := hzServiceToFilter[t.ServiceName]
+		if !ok {
+			toFilter = strings.TrimPrefix(toFilter, "hz:impl:")
+			toFilter = strings.TrimSuffix(toFilter, "Service")
+		}
```
yuce, line 115: How about using this logic for all data structures, and ditch `hzServiceToFilter` ?

list_objects/list.go
```diff
@@ -0,0 +1,123 @@
...
+		if filter == "" {
+			names = append(names, fmt.Sprintf("%s %s", toFilter, t.Name))
```
yuce, line 117: You can add `continue` after this line to get rid of the `else` branch.

list_objects/list_test.go
```diff
@@ -0,0 +1,116 @@
...
+				fmt.Sprintf("PNCounter %s", figName),
+				fmt.Sprintf("PNCounter %s", pnCounterName),
+				fmt.Sprintf("list %s", listName),
+				fmt.Sprintf("map %s", mapName1),
+				fmt.Sprintf("map %s", mapName2),
+				fmt.Sprintf("multiMap %s", mmapName),
+				fmt.Sprintf("queue %s", queueName),
+				fmt.Sprintf("replicatedMap %s", repMapName),
+				// todo find out why topics output as "replicatedMap"
```
yuce, line 79: There doesn't seem to be a problem with topics.

list_objects/list_test.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, tc := range tcs {
+		var b bytes.Buffer
+		cmd := New(&dc)
+		cmd.SetOut(&b)
+		ctx := context.Background()
+		cmd.SetArgs(tc.args)
+		_, err := cmd.ExecuteContextC(ctx)
+		require.NoError(t, err)
+		out := strings.Split(strings.TrimSpace(b.String()), "\n")
+		// at the of writing this, hazelcast creates an internal replicated map
```
yuce, line 102: "at the time of writing this, Hazelcast"

list_objects/list_test.go
```diff
@@ -0,0 +1,116 @@
...
+		// named "__sql.catalog". The line belows removes that
+		out = filterOutInternalObjects(out)
+		require.Equal(t, tc.expect, out)
+	}
+}
+
+func filterOutInternalObjects(out []string) []string {
+	for i, l := range out {
+		if strings.Contains(l, "__") {
```
yuce, line 111: I think `strings.HasPrefix` would be better.

yuce (changes requested):
Looks good, a few ideas.

list_objects/list.go
```diff
@@ -0,0 +1,123 @@
+package list_objects
```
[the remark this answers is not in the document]
yuce: It's OK to use something like `listobjectscmd` until we move it to the commands.

list_objects/list.go
utku-caglayan: I think it wouldn't work since `else if` branch makes an additional check

list_objects/list_test.go
utku-caglayan: Thanks @yuce, keeping this open until the Go client PR is merged.

list_objects/list.go
yuce: What I meant was getting rid of the `else`, something like:
```
if filter == "" {
    names = append(names, fmt.Sprintf("%s %s", toFilter, t.Name))
    continue
}
if filter == toFilter {
    names = append(names, t.Name)
}
```

list_objects/list_test.go
utku-caglayan: Because of flow of the test this is used, `__` is not always the prefix.
case_1: `typeName objectName`
case_2: `objectName`

list_objects/list_test.go
utku-caglayan: Thanks! https://github.com/hazelcast/hazelcast-commandline-client/pull/133/commits/3bb510a54ce0d7b155bd265e343a7f2089c7bbc4

list_objects/list_test.go
utku-caglayan: There was a  bug in the test case. You are right https://github.com/hazelcast/hazelcast-commandline-client/pull/133/commits/3bb510a54ce0d7b155bd265e343a7f2089c7bbc4

list_objects/list.go
utku-caglayan: Refactored for clarity as suggested https://github.com/hazelcast/hazelcast-commandline-client/pull/133/commits/a080288d2f9078ce51c3b47c6b8125e65f10bec5

list_objects/list.go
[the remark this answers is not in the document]
utku-caglayan: https://github.com/hazelcast/hazelcast-commandline-client/commit/3bb510a54ce0d7b155bd265e343a7f2089c7bbc4

list_objects/list.go
utku-caglayan: Sounds good https://github.com/hazelcast/hazelcast-commandline-client/pull/133/commits/e0fb36aefac300d33e2b922bdd99f3ba56a48af6

list_objects/list_test.go
utku-caglayan: Go client fix update is here https://github.com/hazelcast/hazelcast-commandline-client/pull/133/commits/c34809fe0c997fb10c232bba01701bbf734a7ae0

list_objects/list_test.go
yuce: I thought _sql.catalog was the only internal name, are there others?

utku-caglayan: It's the only one. I tried to describe the structure of the strings that may come in `out` parameter.

yuce (approved):

PR #136, opened 2022-09-12
Additions to map commands [API-1540]

utku-caglayan:
This work adds the following map commands to CLC

- remove many
- keys
- values
- entries
- size
- destroy
- lock
- tryLock
- unLock (only for interactive mode)
- set

Also with the addition of "set" command, the put command now prints the older value of the corresponding key


utku-caglayan:
Dear reviewer, please also review the documentation updates on https://github.com/hazelcast/hz-docs/pull/522

Issue #111, opened 2022-08-02
"sql -a non-existing-member" command does not give any warning [API-1563]


feysahin:
**Reproducer:**

When only one member (127.0.0.1:5701) exists, I enter `sql -a 127.0.0.1:5702` after entered `./hzc`, I can start the SQL browser without any warning about the non-existent member.

yuce:
This is not applicable anymore, closing the ticket.

Issue #110, opened 2022-08-02
Invalid commands are silently ignored


feysahin:
When I enter an invalid command with `./hzc` or `sql`, it keeps running without even a warning. This might mislead the user.

**Reproducer:**

- I enter `./hzc -deneme`, and I can connect to the cluster without any warning about the invalid string "deneme".

yuce:
This is not applicable anymore, closing the ticket.

PR #144, opened 2022-09-23
Fix Flag Validation for Interactive Mode [API-1563]

utku-caglayan:
Fix flag validation for interactive mode (turns out it works ok for non-interactive)

Also apart from that, CLC-related errors were being logged instead of "being printed to stderr". This PR fixes that.


utku-caglayan:
fixes #110 and fixes #111

yuce (dismissed):

yuce (approved):

PR #145, opened 2022-09-26
Assigns Viridian as the default coordinator for cloud invocations [API-1556]

utku-caglayan:


run.go
```diff
@@ -121,9 +124,20 @@ func updateConfigWithFlags(rootCmd *cobra.Command, cnfg *config.Config, programA
 	_ = subCmd.ParseFlags(flags)
 	// initialize config from file
 	err := config.ReadAndMergeWithFlags(globalFlagValues, cnfg)
+	setDefaultCoordinator(cnfg.Logger)
 	return err
 }
 
+func setDefaultCoordinator(logger *log.Logger) {
```
yuce, line 131: Why does this function get a logger as parameter?

yuce:
@utku-caglayan Could you also change:
```
--cloud-token string    your Hazelcast Cloud token
```
to:
```
--cloud-token string    your Hazelcast Viridian token
```

run.go
utku-caglayan: It logs if it can not assign the default URL. Do you suggest an alternative flow?

run.go
```diff
@@ -121,9 +124,20 @@ func updateConfigWithFlags(rootCmd *cobra.Command, cnfg *config.Config, programA
...
+	if os.Getenv("HZ_CLOUD_COORDINATOR_BASE_URL") != "" {
```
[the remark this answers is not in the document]
utku-caglayan: This and `--cloud-token` description is fixed here https://github.com/hazelcast/hazelcast-commandline-client/pull/145/commits/0508b2e40ebb6913d3b67f1281f9c0e4303d0e5b

run.go
yuce: Maybe you can just return the error. Since if the coordinator URL cannot be set we shouldn't continue to run the command anyway.

run.go
```diff
@@ -36,6 +37,11 @@ import (
 	"github.com/hazelcast/hazelcast-commandline-client/types/mapcmd"
 )
 
+const (
+	ViridianCoordinatorURL    = "https://api.viridian.hazelcast.com"
+	HzCloudCoordinatorBaseUrl = "HZ_CLOUD_COORDINATOR_BASE_URL"
```
yuce, line 42: Could you make `Url` all-uppercase?

yuce: How about EnvHzCloudCoordinatorBaseURL ?

utku-caglayan: https://github.com/hazelcast/hazelcast-commandline-client/pull/145/commits/8738eaeded94567646916d6707a5ad0a089612a8

run.go
utku-caglayan: I think for many usecases this may be non-critical. For those cases just logging makes sense for me. Wdyt?

yuce: I think it's a serious problem if the user intends to connect to Viridan but it can't since the coordinator URL was not set correctly. What we can do is to call `setDefaultCoordinator` only when we know the user wants to connect to the cloud.

utku-caglayan: Good idea, updated as discussed https://github.com/hazelcast/hazelcast-commandline-client/pull/145/commits/e51609e62016f69e0a9e8fbe2c7e3617f471846d

yuce (approved):
Looks good!

PR #149, opened 2022-10-06
Command reference docs

Serdaro:


emreyigit:
I see documentation of the `force-unlock` but `unlock` which is also a valid command. Could you also add that?

Serdaro:
@emreyigit , added.

JakeSCahill (approved):

PR #155, opened 2022-10-13
Fix spinner text "Index out of range" error

mehmettokgoz:
When the addresses section is empty in the config file (default behavior when connecting to Viridian), the spinner throws out of range error since it is trying to get the first address in the config. Also, it is not correct since there might be more than one member, and we could connect another member differently than we display in the spinner.


yuce:
Changed the spinner code, closing this PR.

PR #156, opened 2022-10-13
Prevent localhost address on config when Viridian is enabled.

mehmettokgoz:
Currently, config template writes `localhost:5701` when there is no address specified, but it is inconvenient since user might be connecting to Viridian. This PR fixes and leave addresses empty when cloud is enabled.


yuce:
We've changed the configuration format, so closing this PR.

PR #160, opened 2022-10-31
Refactoring 1 Plugin interface [API-1648]

yuce:
This PR includes the plugin interfaces, the logger implementation and some utilities.


Makefile
```diff
@@ -2,7 +2,7 @@
 
 GIT_COMMIT=$(shell git rev-parse HEAD 2> /dev/null || echo unknown)
 CLC_VERSION=$(shell git describe --tags `git rev-list --tags --max-count=1` || echo UNKNOWN)
-LDFLAGS="-X 'github.com/hazelcast/hazelcast-go-client/internal.ClientType=CLC' -X 'github.com/hazelcast/hazelcast-commandline-client/internal.GitCommit=$(GIT_COMMIT)' -X 'github.com/hazelcast/hazelcast-commandline-client/internal.ClientVersion=$(CLC_VERSION)' -X 'github.com/hazelcast/hazelcast-go-client/internal.ClientVersion=$(CLC_VERSION)'"
+LDFLAGS="-X 'github.com/hazelcast/hazelcast-go-client/internal.ClientType=CLC' -X 'github.com/hazelcast/hazelcast-commandline-client/internal.GitCommit=$(GIT_COMMIT)' -X 'github.com/hazelcast/hazelcast-commandline-client/internal.Version=$(CLC_VERSION)' -X 'github.com/hazelcast/hazelcast-go-client/internal.ClientVersion=$(CLC_VERSION)'"
```
mehmettokgoz, line 5: Is it valid to replace `ClientVersion` with `Version`? We have `ClientVersion` variable in `internal/release.go` file and I thought this line set up this variable. https://github.com/yuce/hazelcast-commandline-client/blob/refactor-stage-1/internal/release.go#L6

Also, we are using this variable in `version` command so we need to refactor it as well if we select `Version`.

go.mod
```diff
@@ -59,3 +66,5 @@ require (
 	// termdbms
 	github.com/muesli/reflow v0.3.0
 )
+
+replace github.com/hazelcast/hazelcast-go-client v1.3.1-0.20221005090358-58f17bb84ba8 => github.com/yuce/hazelcast-go-client v1.1.2-0.20221024101337-f9de25ad091f
```
mehmettokgoz, line 70: Should be removed.

clc/paths/paths.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func Home() string {
+	dir := os.Getenv(EnvCLCHome)
+	if dir == "" {
+		cd, err := os.UserHomeDir()
+		if err != nil {
+			panic(err)
+		}
+		dir = filepath.Join(cd, unixDir)
```
mehmettokgoz, line 23: As far as I remember, we did research about directories on different OS and in here we are using the UNIX style. Will it be changed later for Windows?

clc/paths/paths.go
```diff
@@ -0,0 +1,87 @@
...
+	return filepath.Join(Home(), "schemas")
+}
+
+func Logs() string {
+	return filepath.Join(Home(), "logs")
+}
+
+func DefaultConfigPath() string {
+	return filepath.Join(Configs(), "default", "config.yaml")
+}
```
mehmettokgoz, line 42: We have only one default config, is it a good practive to create another folder for default?

clc/paths/paths_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/hazelcast/hazelcast-commandline-client/clc/paths"
+	. "github.com/hazelcast/hazelcast-commandline-client/internal/check"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/it/skip"
+)
+
+func TestHomeDir_Unix(t *testing.T) {
+	skip.If(t, "os = windows")
```
mehmettokgoz, line 17: Is it possible to add tests for Windows?

clc/paths/paths.go
```diff
@@ -0,0 +1,87 @@
...
+
+func ResolveLogPath(path string) string {
+	if path == "" {
+		return DefaultLogPath(time.Now())
+	}
+	return path
+}
+
+func Join(paths ...string) string {
```
mehmettokgoz, line 74: We already have `filepath.Join()`, why do we need this function?

clc/paths/paths_test.go
```diff
@@ -0,0 +1,78 @@
+package paths_test
```
mehmettokgoz, line 1: Tests for `ResolveLogPath`, `ResolveConfigPath`, `Schemas` are missing, maybe we can test them too to increase coverage. Ignore this if you think it is unnecessary.

errors/error.go
```diff
@@ -47,6 +47,11 @@ For operations that change state/configuration of the cluster (e.g. "cluster cha
 To change CLUSTER_WRITE permission, see the documentation: https://docs.hazelcast.com/hazelcast/latest/maintain-cluster/rest-api#using-rest-endpoint-groups`
 )
 
+var (
+	ErrUserCancelled = errors.New("user cancelled")
+	ErrNotDecoded    = errors.New("not decoded")
```
mehmettokgoz, line 52: I couldn't find any occurrences for these two error definitions. Will they be used later on in refactoring stage two or three or should we remove them?

internal/output/const.go
```diff
@@ -0,0 +1,15 @@
...
+	TypeStringPretty  = "pretty"
+	TypeStringCSV     = "csv"
+	TypeStringJSON    = "json"
+)
+
+const (
+	ValueUnknown    = "UNKNOWN"
+	ValueNotDecoded = "*"
+	ValueSkip       = ">"
+	ValueNil        = "-"
```
mehmettokgoz, line 14: IMO, instead of showing `-` to user, `nil`or `null` makes more sense.

Makefile
yuce: Since the PRs are coming one by one for the refactor, there may be initial inconsistencies. This is what we are aiming to go: https://github.com/yuce/hazelcast-commandline-client/blob/refactor/internal/version.go

clc/const.go
```diff
@@ -0,0 +1,21 @@
+package clc
+
+const (
+	ShortcutConfigPath     = "c"
+	PropertyClusterAddress = "cluster.address"
```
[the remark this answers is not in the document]
yuce: These are supposed to be keys in the common "properties".

go.mod
yuce: This is currently required for compact support. I will later point that to a branch in the main repo.

clc/paths/paths.go
yuce: In the new configuration scheme, there one or more configuration directories under `$CLC_HOME/configs`, like:
```
$ tree -d `clc home configs`
/home/yuce/.local/share/clc/configs
├── default
├── pr-1272
├── pr-3066
└── _pr-3066.BAC
```
`default/config.yaml` is the default configuration.

clc/paths/paths.go
yuce: `filepath.Join` doesn't work as required if the second path is an absolute path. In this code: https://go.dev/play/p/93q3FUZ4Br2 it prints `/foo/bar/foo/qoo` instead of `/foo/qoo`

errors/error.go
yuce: RIght, they will be used in the serialization related PR.

internal/output/const.go
yuce: We can later change this if necessary.

internal/output/output.go
```diff
@@ -0,0 +1,57 @@
+package output
+
+const (
+	NameKey       = "__key"
+	NameKeyType   = "__key_type"
```
mehmettokgoz, line 5: Possible typo `__key.type`? If it intentional, ignore this comment.

internal/mk/types.go
```diff
@@ -0,0 +1,7 @@
+package mk
+
+import "github.com/hazelcast/hazelcast-commandline-client/internal"
+
+func ValueFromString(value, valueType string) (any, error) {
+	return internal.ConvertString(value, valueType)
```
mehmettokgoz, line 6: Why not use `internal.CovertString()` directly? 🤔

internal/output/ordered_set.go
```diff
@@ -0,0 +1,53 @@
+package output
+
+type OrderedSet[T comparable] struct {
```
mehmettokgoz, line 3: OrderedSet type does not have `Get(index int)` method. It would be good to add this too. If it is not useful for us, ignore this comment.

mehmettokgoz:
Thanks, just a few suggestions. Overall, the change looks very nice but I guess we will see the working architecture when refactor is completed, so skipped some of the missing details. 🙌

mehmettokgoz (commented):
Thanks, just a few suggestions. Overall, the change looks very nice but I guess we will see the working architecture when refactor is completed, so skipped some of the missing details. 🙌

internal/output/output.go
yuce: That's intentional.

internal/mk/types.go
yuce: I'll probably remove that function.

internal/output/ordered_set.go
yuce: Let's add it when necessary.

clc/paths/paths.go
yuce: Added that at: https://github.com/hazelcast/hazelcast-commandline-client/pull/160/commits/b9fd102d81c6100a8d5beef5b705fbf6f42f464e

clc/paths/paths_test.go
yuce: Added Windows tests at: https://github.com/hazelcast/hazelcast-commandline-client/pull/160/commits/b9fd102d81c6100a8d5beef5b705fbf6f42f464e Note that these should be run manually ATM.

clc/paths/paths.go
yuce: Added that at: https://github.com/hazelcast/hazelcast-commandline-client/pull/160/commits/b9fd102d81c6100a8d5beef5b705fbf6f42f464e

clc/paths/paths_test.go
yuce: Added more tests at: https://github.com/hazelcast/hazelcast-commandline-client/pull/160/commits/8b93697886f93b70a5932dda90559e8d4d075f35

internal/mk/types.go
yuce: Clarification: I'll probably remove `internal.ConvertString` in the future.

mehmettokgoz (approved):
LGTM!

PR #161, opened 2022-10-31
Refactoring 2 [API-1649]

yuce:
This PR adds the initial CLC command and implements the `plug.CommandContext` and `plug.ExecContext` interfaces.

The plugins for actual commands will be sent in a separate PR.


clc/cmd/command_context.go
```diff
@@ -0,0 +1,126 @@
...
+	return cc.isInteractive
+}
+
+func (cc *CommandContext) SetCommandHelp(long, short string) {
+	if long != "" {
+		cc.Cmd.Long = long
+	}
+	if short != "" {
+		cc.Cmd.Short = short
+	}
```
emreyigit, line 87: What if both arguments are empty?

clc/cmd/command_context.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/hazelcast/hazelcast-commandline-client/internal/check"
+	. "github.com/hazelcast/hazelcast-commandline-client/internal/check"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/mk"
+)
+
+type CommandContext struct {
+	Cmd           *cobra.Command
+	Vpr           *viper.Viper
+	stringValues  map[string]*string
```
emreyigit, line 15: Just wondering. What's the value of `*string` or same for other primitives in the map?

yuce: Using primitives would copy their values, so Cobra wouldn't be able to change them.

clc/cmd/command_context.go
yuce: It wouldn't display the help string for that command.

clc/cmd/exec_context.go
```diff
@@ -0,0 +1,194 @@
...
+	if ec.ci != nil {
+		return ec.ci, nil
+	}
+	client, err := ec.ExecuteBlocking(ctx, "Connecting to the cluster", func(ctx context.Context) (any, error) {
+		return ec.clientFn(ctx)
+	})
+	if err != nil {
+		return nil, err
+	}
+	ec.ci = hazelcast.NewClientInternal(client.(*hazelcast.Client))
```
emreyigit, line 101: I couldn't understand here. What's happening?

emreyigit (approved):

PR #162, opened 2022-10-31
Refactoring 3

yuce:
TODO


internal/serialization/portable_field.go
```diff
@@ -0,0 +1,380 @@
...
+	case PortableTypeTimestampWithTimezoneArray:
+		return ValueNotDecoded
+	default:
+		return ValueUnknown
+	}
+	return v
+}
+
+// PortableFieldType corresponds to FieldDefinitionType+1
+type PortableFieldType int32
```
mehmettokgoz, line 98: Why we are not using FieldDefinitionType directly from Go Client?

internal/serialization/portable_field.go
```diff
@@ -0,0 +1,380 @@
...
+	case PortableTypeInt64:
+		s = "int64"
+	case PortableTypeFloat32:
+		s = "float32"
+	case PortableTypeFloat64:
+		s = "float64"
+	case PortableTypeString:
+		s = "string"
+	case PortableTypePortableArray:
+		s = "portablearray"
```
mehmettokgoz, line 160: Java client uses uppercase notation for these strings with `_` between words. I think we should follow same here.

internal/serialization/generic_compact.go
```diff
@@ -0,0 +1,131 @@
...
+			continue
+		}
+		v.Elem().Field(i).Set(value)
+	}
+	return v.Interface()
+}
+
+func (cm GenericCompact) Write(writer serialization.CompactWriter, value interface{}) {
+	// TODO: implement me
+	panic("implement me")
```
mehmettokgoz, line 81: More revealing explanation can be given like `write operation for compact serialization is not supported yet.`

internal/output/utils.go
```diff
@@ -0,0 +1,93 @@
...
+	// delete standard headers
+	for _, h := range stdNames {
+		hd.Delete(h)
+	}
+	header := hd.Items()
+	/*
+		sort.Slice(header, func(i, j int) bool {
+			return header[i] < header[j]
+		})
+	*/
```
mehmettokgoz, line 71: Should we delete this?

internal/serialization/compact_field.go
```diff
@@ -0,0 +1,47 @@
...
+	Value any
+}
+
+var compactReaders = map[serialization.FieldKind]compactFieldReader{
+	serialization.FieldKindString: func(r serialization.CompactReader, field string) any {
+		v := r.ReadString(field)
+		if v == nil {
+			return nil
+		}
+		return *v
```
mehmettokgoz, line 22: Isn't this logic suppose to be inside the `ReadString`, why do we explicitly edit the return? I see the same operation in https://github.com/yuce/hazelcast-go-client/blob/compact-serialization-temp/internal/serialization/default_compact_reader.go#L224.

internal/output/column.go
```diff
@@ -1,7 +1,352 @@
...
+}
+
+func (co Column) SingleLine() (s string) {
+	/*
+		defer func() {
+			if e := recover(); e != nil {
+				s = ValueNotDecoded
+			}
+		}()
+	*/
```
mehmettokgoz, line 75: Should we remove this?

internal/output/column.go
```diff
@@ -1,7 +1,352 @@
...
+)
+
 type Column struct {
 	Name  string
 	Type  int32
 	Value any
 }
+
+func NewStringColumn(value string) Column {
+	return Column{Type: iserialization.TypeString, Value: value}
```
mehmettokgoz, line 25: Can we follow the same styling (multiple line) with others?

internal/serialization/portable_field.go
yuce: `FieldlDefinitionType` starts from `0` and it means `TypePortable`. The zero value for `int32` is also `0`, so when used in `PortableField` and not setting the `Type` field, it would mean that portable field contains portable data.

internal/serialization/generic_compact.go
yuce: That code is never executed by CLC. I think it's purpose would be probably understood by developers.

internal/serialization/compact_field.go
yuce: We want to return a `string` here, not `*string`.

internal/serialization/portable_field.go
yuce: I guess it's a matter of preference. We can discuss that later, if it's OK with you.

internal/output/utils.go
yuce: Removed at: https://github.com/hazelcast/hazelcast-commandline-client/pull/162/commits/e7f4a6f200af7ce276b0b087c2f91f4340ef8140

internal/output/column.go
yuce: Uncommented at: https://github.com/hazelcast/hazelcast-commandline-client/pull/162/commits/e7f4a6f200af7ce276b0b087c2f91f4340ef8140

mehmettokgoz (approved):
Looks good, thanks!

PR #165, opened 2022-11-24
Refactor 2 [API-1751] [API-1752]

yuce:


rc.sh
```diff
@@ -170,7 +170,7 @@ help () {
 
 TIMESTAMP_FMT="+%Y-%m-%d %H:%M:%S"
 PID_FILE="test.pid"
-HZ_VERSION="${HZ_VERSION:-5.1-SNAPSHOT}"
+HZ_VERSION="${HZ_VERSION:-5.2.1}"
```
harunalpak, line 173: Should we use "5.2.2-SNAPSHOT" or not? I am not sure.

internal/output/utils.go
```diff
@@ -75,13 +87,126 @@ func MakeTable(rows []Row) ([]string, []Row) {
...
+	for i, c := range row {
+		hd[i] = table.Column{
+			Header: c.Name,
+			Align:  alignmentForType(c.Type) * widthForType(c.Type, len(c.Name), dw),
+		}
+	}
+	return hd
+}
+
+func alignmentForType(t int32) int {
```
harunalpak, line 133: For this function may be we can use enum for easy reading code.

```
const (
	LeftAlign    int = -1
	RightAlign  int = 1
)
```

internal/output/utils.go
```diff
@@ -75,13 +87,126 @@ func MakeTable(rows []Row) ([]string, []Row) {
...
+			Header: strings.TrimPrefix(h, thisPfx),
+			Align:  align[i] * width[i],
+		}
+	}
+	return columns, rows
+}
+
+func makeTableHeaderFromRow(row Row, maxWidth int) table.Row {
+	dw := defaultWidthForFlexibleColumns(row, maxWidth)
+	// the max width for a column is 1000, just an arbitrary decision --YT
```
harunalpak, line 115: We have used max width only for header. Maybe we can set this value lower and also we have to use for both header and also column the same value.

internal/output/utils.go
```diff
@@ -75,13 +87,126 @@ func MakeTable(rows []Row) ([]string, []Row) {
...
+			fixedSize += widthForType(h.Type, len(h.Name), len(h.Name)) + padding
+		default:
+			flexibleCount++
+		}
+	}
+	if flexibleCount == 0 {
+		return 1
+	}
+	w := (maxWidth-fixedSize)/flexibleCount - padding
+	if w < 0 {
```
harunalpak, line 190: When we return 1, is this means that for this column we only show it 1 width? I did not understand what will show if it is 1.
Maybe we can set a lower limit, if the value is lower than this we may not show this column.

internal/output/table_result.go
```diff
@@ -2,87 +2,68 @@ package output
...
-	header []string
-	rp     RowProducer
+	header   []table.Column
+	rp       RowProducer
+	maxWidth int
 }
 
-func NewTableResult(header []string, rp RowProducer) *TableResult {
+// NewTableResult creates a table result from the row producer and (optional) header.
+// If header is not given, then it is assumed the first row in the row producer is the header, and alignment is auto-calculatd.
```
harunalpak, line 20: Typo-fix 
`auto-calculatd.=> auto-calculated.`

internal/output/utils.go
yuce: The columns use the same width/alignment with the header. This code is used only when the table is in "streaming mode" (the max width of columns not found) as a safety measure.

internal/output/utils.go
yuce: > When we return 1, is this means that for this column we only show it 1 width?
That's right. 

> Maybe we can set a lower limit, if the value is lower than this we may not show this column.
I have a different idea about hiding columns, something like: "display only x columns, and display `...` for the rest" if there's no way to fit the column to the width (console or set by user). But that's not implemented yet.

internal/output/table_result.go
yuce: Fixed at: https://github.com/hazelcast/hazelcast-commandline-client/pull/165/commits/04a0368bd33191940eb258a57ba49b5ea45d97c9

internal/output/utils.go
yuce: Added comments at: https://github.com/hazelcast/hazelcast-commandline-client/pull/165/commits/04a0368bd33191940eb258a57ba49b5ea45d97c9 Since this is a multipler, not sure making them enums would be beneficial.

harunalpak (approved):

PR #176, opened 2022-12-08
Readline update

yuce:
* Adds `go-readline-ny` as an experimental readline library.
* Limits error output in the shell
* Fixes the interactive mode on Windows*


clc/shell/linereader.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	formatter := os.Getenv(envFormatter)
+	if formatter == "" || !strings.HasPrefix(formatter, "terminal") {
+		formatter = "terminal"
+	}
+	cmdColor := color.New(color.Bold)
+	cfg := &gohxs.Config{
+		Prompt:          prompt,
+		HistoryFile:     sh.historyPath,
+		InterruptPrompt: "^C",
```
harunalpak, line 86: Ctrl+C generally used for exit. Why we interrupt this command?

PR #188, opened 2023-03-02
Add binary installation instructions.

Serdaro:


angelasimms (dismissed):
Looks good

docs/modules/ROOT/pages/install-clc.adoc
```diff
@@ -40,14 +40,24 @@ cd hazelcast-commandline-client
...
 
+The `clc` binary is created in the `build` directory.
 . Run the following command to start the Hazelcast CLC.
 +
 [source,shell]
 ----
 ./build/clc
 ----
+
+ZIP (AMD)::
```
yuce, line 52: How about `AMD64` just to use the correct arch name?

yuce (approved):
Thanks, looks good! minor point, but OK anyway

Issue #186, opened 2023-02-28
Map name parameter gets stuck to a value in the shell mode


yuce:
Reproducer:

```
CLC> \map set foo bar
CLC> \map size
 Size 
    1 

CLC> \map size -n othermap
 Size 
    1 
```

PR #189, opened 2023-03-10
Map name parameter gets stuck to a value in the shell mode [API-1934]

mehmettokgoz:
Fixes: #186 

Blocking call of `map` was saving the created proxy object and returning it in the subsequent commands directly without creating a new proxy. This behaviour results in constant map proxy even if a different map name is specified. 

A new check is added to return if the saved proxy's name matches with the requested map name.


mehmettokgoz:
Alternatively, we can use a map with mutex to store the proxy object for each map requested previously and return it. Wdyt @yuce?

yuce:
@mehmettokgoz How about just removing the caching of the map object? The client already caches the map, so no need to cache it again.

Could you try removing `m` from the `MapCommand`?
```
type MapCommand struct {
	m *hazelcast.Map
}
```

yuce (approved):

PR #203, opened 2023-04-05
Refactored Output

yuce:
* Refactored stringifying data
* Added more tests


Makefile
```diff
@@ -17,10 +17,10 @@ build:
 	CGO_ENABLED=0 go build -tags base,hazelcastinternal,hazelcastinternaltest -ldflags $(LDFLAGS)  -o build/$(BINARY_NAME) ./cmd/clc
 
 test:
-	go test -tags base,hazelcastinternal,hazelcastinternaltest $(TEST_FLAGS) ./...
+	go test -tags base,hazelcastinternal,hazelcastinternaltest -p 1 $(TEST_FLAGS) ./...
```
mehmettokgoz, line 20: Is this `-p` flag related to parallelism?

.github/workflows/coverage_runner.yaml
```diff
@@ -65,10 +77,11 @@ jobs:
           
       - name: Test
         env:
-          HAZELCAST_ENTERPRISE_KEY: ${{ secrets.HAZELCAST_ENTERPRISE_KEY }}
+#          HAZELCAST_ENTERPRISE_KEY: ${{ secrets.HAZELCAST_ENTERPRISE_KEY }}
```
mehmettokgoz, line 80: Why don't

.github/workflows/coverage_runner.yaml
```diff
@@ -65,10 +77,11 @@ jobs:
...
           HZ_VERSION: 5.2.1
-          SSL_ENABLED: 1
+#          SSL_ENABLED: 1
```
mehmettokgoz, line 82: Why don't we test SSL enabled?

internal/it/expect/expect.go
```diff
@@ -60,13 +60,15 @@ func (e *Expect) String() string {
...
 func (e *Expect) Match(m Matcher, options ...Option) bool {
-	//e.Reset()
 	o := Options{}
 	for _, opt := range options {
 		if err := opt(&o); err != nil {
 			panic(fmt.Errorf("creating Match options: %w", err))
 		}
 	}
+	if o.delay > 0 {
+		time.Sleep(o.delay)
```
mehmettokgoz, line 70: Show we multiply with seconds here, or expect it to come as duration object?

internal/serialization/portable_serialization_it_test.go
```diff
@@ -0,0 +1,217 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, tc.f)
+	}
+}
+
+func portablePrimitivesTest(t *testing.T) {
+	portableTester(t, func(tcx it.TestContext) {
+		value := &primitives2{
+			valueByte: 8,
+			valueBool: true,
```
mehmettokgoz, line 35: `int` types are missing for portable tests.

internal/serialization/types.go
```diff
@@ -1,3 +1,232 @@
...
+	}
+	sb.WriteString("]")
+	return sb.String()
+}
+
+func javaClassStringer(v any) string {
+	return fmt.Sprintf("Java Class: %v", v)
+}
+
+func timeStringer(v any) string {
```
mehmettokgoz, line 135: Didn't we added this functionality to Go client? Why don't we use it there?

Also instead of creating these stringer functions and registering to a map I think we should bound these methods to types and call `ToString()` directly.

internal/serialization/stringer_test.go
```diff
@@ -0,0 +1,97 @@
...
+		{
+			name:   "*types.OffsetDateTime",
+			in:     (*types.OffsetDateTime)(&tm),
+			target: "2023-01-02T03:04:05Z",
+		},
+		{
+			name:   "types.OffsetDateTime",
+			in:     (types.OffsetDateTime)(tm),
+			target: "2023-01-02T03:04:05Z",
+		},
```
mehmettokgoz, line 88: Can you add a test for non-UTC offset here? Also the formatting does not match IMO:

`LocalDateTime`: 2023-01-02 03:04:05
`OffsetDateTime`: 2023-01-02T03:04:05Z

internal/serialization/compact_field.go
```diff
@@ -67,37 +59,37 @@ var compactReaders = map[serialization.FieldKind]compactFieldReader{
...
 	serialization.FieldKindTime: func(r serialization.CompactReader, field string) any {
-		return readNullableCompactField(field, r.ReadTime)
+		return r.ReadTime(field)
 	},
 	serialization.FieldKindArrayOfTime: func(r serialization.CompactReader, field string) any {
 		return r.ReadArrayOfTime(field)
 	},
 	serialization.FieldKindDate: func(r serialization.CompactReader, field string) any {
-		return readNullableCompactField(field, r.ReadDate)
+		return r.ReadDate(field)
```
mehmettokgoz, line 80: AFAIS we removed nullable tag here but some of the other methods still contains it. Will we have different implementations for nullable?

internal/serialization/generic_compact.go
```diff
@@ -19,109 +14,74 @@ type SchemaInfo struct {
...
-		fds := schema.FieldDefinitions()
-		fs := make([]reflect.StructField, len(fds))
-		for i, f := range fds {
-			fs[i] = reflect.StructField{
-				Name: fmt.Sprintf("Field%03d", i),
-				Type: fieldKindToType[f.Kind],
-				Tag:  reflect.StructTag(fmt.Sprintf("json:\"%s\"", f.Name)),
-			}
+		if v == nil {
+			c.Type = TypeNil
```
mehmettokgoz, line 30: We already have `FieldKindNotAvailable` type, isn't this available through `fd.kind`?

base/commands/config/config_it_test.go
```diff
@@ -0,0 +1,67 @@
...
+		{name: "Add", f: addTest},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, tc.f)
+	}
+}
+
+func importTest(t *testing.T) {
+	tcx := it.TestContext{T: t}
+	const configURL = "https://rcd-download.s3.us-east-2.amazonaws.com/hazelcast-cloud-go-sample-client-pr-FOR_TESTING-default.zip"
```
mehmettokgoz, line 31: Should we add this as secret to repo? I checked there is no configuration :).

internal/output/column.go
```diff
@@ -1,25 +1,10 @@
...
-	"github.com/hazelcast/hazelcast-commandline-client/errors"
 	iserialization "github.com/hazelcast/hazelcast-commandline-client/internal/serialization"
 )
 
-type Column struct {
-	Name  string
-	Type  int32
-	Value any
-}
+type Column = iserialization.Column
```
mehmettokgoz, line 7: Why do we implement special methods for `string`, `nil` and `skip`? Can't we use NewValueColumn since already we are using it for most of the types?

yuce: Removed helpers at: https://github.com/hazelcast/hazelcast-commandline-client/pull/203/commits/d1951c4d77fe39388b506e7553f5776a5fc09271

mehmettokgoz (approved):
Thanks, LGTM!

PR #221, opened 2023-04-28
Add "map destroy" command [CLC-154]

mtyazici:


base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/hazelcast/hazelcast-go-client"
+
+	"github.com/hazelcast/hazelcast-commandline-client/clc"
+	. "github.com/hazelcast/hazelcast-commandline-client/internal/check"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/plug"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/prompt"
+)
+
+const (
+	mapDestroyFlagAutoYes = "yes"
```
yuce, line 17: Could you move this to `clc/const.go`? We will use it with other dangerous commands too.

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type MapDestroyCommand struct{}
+
+func (mc *MapDestroyCommand) Init(cc plug.InitContext) error {
+	help := "Destroy a Map"
+	cc.SetCommandHelp(help, help)
```
yuce, line 24: Maybe we can add more information in the long help. Something like:
```
Destroy a Map

This command will delete the Map and the data in it will not be available anymore.
```

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+	cc.AddBoolFlag(mapDestroyFlagAutoYes, "", false, false, "skip interactive approval")
```
yuce, line 25: How about "skip conforming the destroy operation"

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+	cc.SetCommandUsage("destroy")
+	return nil
+}
+
+func (mc *MapDestroyCommand) Exec(ctx context.Context, ec plug.ExecContext) error {
+	mv, err := ec.Props().GetBlocking(mapPropertyName)
+	if err != nil {
+		return err
+	}
+
```
yuce, line 35: Could you remove the blank line?

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+	autoYes := ec.Props().GetBool(mapDestroyFlagAutoYes)
+	if !autoYes {
+		prompt := prompt.NewPrompter(ec.Stdin(), ec.Stdout())
+		yes, err := prompt.YesNoPrompt("Map will be deleted irreversibly, do you agree?")
```
yuce, line 39: How about? `Map will be deleted irreversibly, proceed?`

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			return err
```
yuce, line 41: What kind of error we get here?

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+		if !yes {
+			return errors.New("User did not agree")
```
yuce, line 44: Could you return `errors.ErrUserCancelled` instead?

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}
+
```
yuce, line 47: Could you remove the blank line?

base/commands/map/map_destroy.go
```diff
@@ -0,0 +1,65 @@
...
+func (mc *MapDestroyCommand) Init(cc plug.InitContext) error {
+	help := "Destroy a Map"
+	cc.SetCommandHelp(help, help)
+	cc.AddBoolFlag(mapDestroyFlagAutoYes, "", false, false, "skip interactive approval")
+	cc.SetCommandUsage("destroy")
+	return nil
+}
+
+func (mc *MapDestroyCommand) Exec(ctx context.Context, ec plug.ExecContext) error {
+	mv, err := ec.Props().GetBlocking(mapPropertyName)
```
yuce, line 31: How about `m, err := ec.Props().GetBlocking(mapPropertyName).(*hazelcast.Map)` to save a line?

base/commands/map/map_it_test.go
```diff
@@ -146,3 +149,49 @@ func size_InteractiveTest(t *testing.T) {
...
+func destroy_NonInteractiveTest(t *testing.T) {
+	it.MapTester(t, func(tcx it.TestContext, m *hz.Map) {
+		t := tcx.T
+		ctx := context.Background()
+		tcx.WithReset(func() {
+			check.Must(m.Set(ctx, "foo", "bar"))
+			require.Equal(t, 1, check.MustValue(m.Size(ctx)))
+			go tcx.WriteStdin([]byte("y\n"))
+			check.Must(tcx.CLC().Execute(ctx, "map", "-n", m.Name(), "destroy"))
+			require.Equal(t, 0, check.MustValue(m.Size(ctx)))
```
yuce, line 162: Instead of checking the size of the map, it would be better to check its existence. You can use ` tcx.Client.GetDistributedObjectsInfo()` to get the list of objects in the cluster and check whether `m.Name()` is in the list.

base/commands/map/map_it_test.go
```diff
@@ -146,3 +149,49 @@ func size_InteractiveTest(t *testing.T) {
...
+
+func destroy_defaultYes_NonInteractiveTest(t *testing.T) {
+	it.MapTester(t, func(tcx it.TestContext, m *hz.Map) {
+		t := tcx.T
+		ctx := context.Background()
+		tcx.WithReset(func() {
+			check.Must(m.Set(ctx, "foo", "bar"))
+			require.Equal(t, 1, check.MustValue(m.Size(ctx)))
+			check.Must(tcx.CLC().Execute(ctx, "map", "-n", m.Name(), "destroy", "--yes"))
+			require.Equal(t, 0, check.MustValue(m.Size(ctx)))
```
yuce, line 175: Same comment about checking the existence of the map.

docs/modules/ROOT/pages/clc-map.adoc
```diff
@@ -270,3 +271,32 @@ Example:
...
+|Parameter|Required|Description|Default
+
+|`--name`, `-n`
+|Optional
+|Name of the map.
+|`default`
+
+|`--yes`
+|Optional
+|Skip interactive approval.
```
yuce, line 299: Same comment about flag help.

internal/prompt/prompt.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (p *Prompter) YesNoPrompt(q string) (bool, error) {
+	prompt := fmt.Sprintf("%s (y/n) ", q)
+	s, err := p.readline(prompt)
+	if err != nil {
+		return false, err
+	}
+	s = strings.TrimSpace(s)
+	s = strings.ToLower(s)
```
yuce, line 33: How about `s = strings.ToLower(strings.TrimSpace(s))` to save a line?

internal/prompt/prompt.go
```diff
@@ -0,0 +1,59 @@
...
+	switch s {
+	case "y", "yes":
+		return true, nil
+	case "n", "no":
+		return false, nil
+	default:
+		return false, errors.New("Invalid input")
```
yuce, line 40: How about interpreting anything other then `y/yes` as no?

internal/prompt/prompt.go
```diff
@@ -0,0 +1,59 @@
...
+		Prompt:                 prompt,
+		InterruptPrompt:        p.InterruptPrompt,
+		Stdout:                 p.Stdout,
+		Stdin:                  p.Stdin,
+		DisableAutoSaveHistory: true,
+		HistoryLimit:           -1,
+	}
+	rl, err := gohxs.NewEx(cfg)
+	if err != nil {
+		return "", err
```
yuce, line 55: Maybe: `return "", fmt.Errorf("creating gohxs readline: %w", err)`

yuce (changes requested):
Here are a few ideas...

internal/prompt/prompt.go
mtyazici: Sure, I will change it to that.

base/commands/map/map_destroy.go
mtyazici: When user gives a wrong input or any error when getting the user input.

yuce: Would it make sense to assume the answer is "no" if there's an error?

mtyazici: The end result is like the answer was no and an error could help us find out if there was a real problem getting user's input.

base/commands/map/map_it_test.go
```diff
@@ -152,46 +153,48 @@ func size_InteractiveTest(t *testing.T) {
...
-		t := tcx.T
 		ctx := context.Background()
 		tcx.WithReset(func() {
-			check.Must(m.Set(ctx, "foo", "bar"))
-			require.Equal(t, 1, check.MustValue(m.Size(ctx)))
 			go tcx.WriteStdin([]byte("y\n"))
 			check.Must(tcx.CLC().Execute(ctx, "map", "-n", m.Name(), "destroy"))
-			require.Equal(t, 0, check.MustValue(m.Size(ctx)))
+			objects := check.MustValue(tcx.Client.GetDistributedObjectsInfo(ctx))
+			check.MustNotOK(objectExists(hz.ServiceNameMap, m.Name(), objects))
```
yuce, line 161: YOu can also use `require.False` in order not to introduce `MustNotOK`.

internal/check/must.go
```diff
@@ -20,7 +20,19 @@ func MustAnyValue[T any](v any, err error) T {
 	return v.(T)
 }
 
-func MustOK[T any](v any, ok bool) T {
+func MustOK(ok bool) {
```
yuce, line 23: Changing this function shouldn't be necessary. Could you revert the change? In tests we can use `require.True` instead.
I don't think it's used anywhere. Let's remove it when we do the cleanup.

yuce (changes requested):

base/commands/map/map_destroy.go
yuce: How about assuming the answer is "no" but logging the error?

docs/modules/ROOT/pages/clc-map.adoc
```diff
@@ -296,7 +296,7 @@ Parameters:
 
 |`--yes`
 |Optional
-|Skip interactive approval.
+|Skip conforming the destroy operation.
```
yuce, line 299: "confirming"

yuce (dismissed):
Looks good!

yuce (approved):

PR #223, opened 2023-05-02
[CLC-155]: Add map key-set command

kutluhanmetin:
- Adds key-set command.


base/commands/map/map_key_set.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		rows = append(rows, row)
+	}
+	if len(rows) > 0 {
+		return ec.AddOutputRows(ctx, rows...)
+	}
+
+	if !ec.Props().GetBool(clc.PropertyQuiet) {
+		I2(fmt.Fprintln(ec.Stdout(), "No entries found."))
+	}
```
yuce, line 67: Could you replace this and the corresponding code at `MapEntrySetCommand` with:
```
ec.PrintlnUnnecessary("No entries found.")
```

yuce (changes requested):
Looks good. Just a minor idea...

yuce (approved):
Looks good!

PR #226, opened 2023-05-05
Viridian Testing Updates

yuce:
- Improves Viridian connection by reporting discovery token problems.
- Added the initial Viridian test system.
- Added the nightly test workflow.


.github/workflows/nightly_tests.yaml
```diff
@@ -0,0 +1,81 @@
...
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+    runs-on: ${{ matrix.os }}
+    env:
+      GOPATH: "${{ github.workspace }}"
+      HZ_VERSION: "5.3.0-BETA-2"
+      DEFAULT_TIMEOUT: "30s"
+      ENABLE_VIRIDIAN: "1"
+      HZ_CLOUD_COORDINATOR_BASE_URL: "https://api.dev2.viridian.hazelcast.cloud"
+      # Note:
+      # VIRIDIAN_API_KEY and VIRIDIAN_API_SECRET are repo secrets
```
mtyazici, line 19: How is the test going to access the secrets if they are not made environment variables? I believe we need to assign secrets to env variables.

base/commands/viridian/viridian_it_test.go
```diff
@@ -0,0 +1,130 @@
...
+		})
+	})
+}
+
+func ensureClusterRunning(ctx context.Context, tcx it.TestContext) string {
+	info, err := tcx.Viridian.CreateCluster(ctx, it.UniqueClusterName())
+	if err != nil {
+		tcx.T.Logf("Ignoring error: %s", err.Error())
+		infos := check.MustValue(tcx.Viridian.ListClusters(ctx))
+		info = infos[0]
```
mtyazici, line 96: If slice is empty this could panic, is this ok?

internal/it/test_cluster.go
```diff
@@ -0,0 +1,301 @@
...
+	return res.Content, nil
+}
+
+func makeUrl(path string) string {
+	path = strings.TrimLeft(path, "/")
+	path = "/" + path
+	return os.Getenv(envAPIBaseURL) + path
+}
+
+func doPost[Req, Res any](ctx context.Context, path, token string, request Req) (res Res, err error) {
```
mtyazici, line 192: I have a question. I guess we did not want to expose the `doPost`, `doGet` etc. functions in `internal/viridian/api.go`.  I am not sure if we intend to change the code in these functions but does it make sense to create a client package that will expose these functions so that the client package can be used in the prod code and tests?

internal/it/test_cluster.go
```diff
@@ -0,0 +1,301 @@
...
+	m, err := json.Marshal(request)
+	if err != nil {
+		return res, fmt.Errorf("creating login payload: %w", err)
```
mtyazici, line 195:
```suggestion
		return res, fmt.Errorf("creating request payload: %w", err)
```

mtyazici (commented):
I have few questions

base/commands/viridian/viridian_it_test.go
yuce: Yes, in the common case the error should be about "adding payment method" since there would already be a running cluster.

internal/it/test_cluster.go
yuce: We could do that, but this is some trivial code that is supposed to be very simple and should be used just to run the tests.. I think it would be OK to duplicate them in order to not create a dependency. This way we will be able make changes to the API in internal/viridian without updating the tests.

.github/workflows/nightly_tests.yaml
yuce: You are right, added them at: https://github.com/hazelcast/hazelcast-commandline-client/pull/226/commits/9eb193af8b9bf70c8552eba888d78a395697113d

internal/it/test_cluster.go
yuce: https://github.com/hazelcast/hazelcast-commandline-client/pull/226/commits/9eb193af8b9bf70c8552eba888d78a395697113d

mtyazici (approved):
LGTM

PR #228, opened 2023-05-09
[CLC-153] [CLC-157] Job command improvements and tests

yuce:
* Added the `--wait` flag for `job` commands
* Retries both sending the metadata and jar upload if there's an error
* Added `job` tests


base/commands/job/job_it_test.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+const jobPath = "testdata/sample-job-1-1.0-SNAPSHOT-all.jar"
+
+func TestJob(t *testing.T) {
+	testCases := []struct {
+		name string
+		f    func(t *testing.T)
+	}{
+		{name: "listNonInteractive", f: listNonInteractiveTest},
```
kutluhanmetin, line 19: in other tests, we applied suffix with `_`. So maybe, you would like to the same for these test names as well.

kutluhanmetin (approved):
LGTM!

PR #232, opened 2023-05-12
Viridian update

yuce:
Waits for the cluster creation and creates a configuration from the cluster.

### DIFF
--- a/base/commands/viridian/common.go
+++ b/base/commands/viridian/common.go
@@ -1,14 +1,27 @@
 package viridian
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"os"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/hazelcast/hazelcast-commandline-client/clc/secrets"
 	"github.com/hazelcast/hazelcast-commandline-client/internal/plug"
 	"github.com/hazelcast/hazelcast-commandline-client/internal/viridian"
+)
+
+const (
+	stateRunning = "RUNNING"
+	stateFailed  = "FAILED"
+)
+
+var (
+	ErrClusterFailed   = errors.New("cluster failed")
+	ErrClusterNotFound = errors.New("cluster not found")
 )
 
 func findToken(apiKey string) (string, error) {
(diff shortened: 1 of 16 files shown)


base/commands/viridian/common.go
```diff
@@ -53,3 +66,39 @@ func getAPI(ec plug.ExecContext) (*viridian.API, error) {
 	}
 	return viridian.NewAPI(string(token)), nil
 }
+
+func waitClusterState(ctx context.Context, ec plug.ExecContext, api *viridian.API, clusterIDOrName, state string) error {
+	for {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		cs, err := api.ListClusters(ctx)
```
mtyazici, line 75: Is there a reason we are using the `list` endpoint instead of `get`?

mtyazici (approved):

base/commands/viridian/common.go
yuce: That would do 2 HTTP requests instead of 1.

mtyazici: Ah, I see we are using cluster id or name.

base/commands/viridian/download_logs_it_test.go
```diff
@@ -21,24 +22,22 @@ func downloadLogs_NonInteractiveTest(t *testing.T) {
...
 func downloadLogs_InteractiveTest(t *testing.T) {
 	viridianTester(t, func(ctx context.Context, tcx it.TestContext) {
 		dir := check.MustValue(os.MkdirTemp("", "log"))
-		defer func() { check.Must(os.RemoveAll(dir)) }()
+		t.Logf("Downloading to directory: %s", dir)
+		//defer func() { check.Must(os.RemoveAll(dir)) }()
 		tcx.WithShell(ctx, func(tcx it.TestContext) {
 			tcx.WithReset(func() {
 				c := createOrGetClusterWithState(ctx, tcx, "")
+				time.Sleep(10 * time.Second)
```
mtyazici, line 37: Instead of putting sleep would it be better to get a cluster with state `RUNNING`? I guess that is the reason for sleep right?

base/commands/viridian/download_logs_it_test.go
```diff
@@ -21,24 +22,22 @@ func downloadLogs_NonInteractiveTest(t *testing.T) {
...
 		})
 	})
 }
 
 func downloadLogs_InteractiveTest(t *testing.T) {
 	viridianTester(t, func(ctx context.Context, tcx it.TestContext) {
 		dir := check.MustValue(os.MkdirTemp("", "log"))
-		defer func() { check.Must(os.RemoveAll(dir)) }()
+		t.Logf("Downloading to directory: %s", dir)
+		//defer func() { check.Must(os.RemoveAll(dir)) }()
```
mtyazici, line 33: Why are we not deleting the temp directory?

base/commands/viridian/viridian_it_test.go
```diff
@@ -152,8 +155,13 @@ func createCluster_InteractiveTest(t *testing.T) {
 			tcx.WithReset(func() {
 				clusterName := it.UniqueClusterName()
 				tcx.WriteStdinf("\\viridian create-cluster --verbose --name %s \n", clusterName)
+				time.Sleep(10 * time.Second)
```
mtyazici, line 159: Why do we need this sleep?

mtyazici (commented):
I have a few questions.

base/commands/viridian/download_logs_it_test.go
yuce: I agree, updated at: https://github.com/hazelcast/hazelcast-commandline-client/pull/232/commits/f8e7230dae1a5e5703035586a34996ccc93a436b

base/commands/viridian/download_logs_it_test.go
yuce: My bad: https://github.com/hazelcast/hazelcast-commandline-client/pull/232/commits/f8e7230dae1a5e5703035586a34996ccc93a436b

mtyazici (approved):

PR #239, opened 2023-05-16
[CLC-67] Updated table printer

yuce:
* Updated table to have multiline rows.
* Non-stream tables observe `CLC_MAX_COLS` environment variable.
* Other refactoring.

### DIFF
--- a/base/printers.go
+++ b/base/printers.go
@@ -55,7 +55,7 @@ type TablePrinter struct{}
 func (pr *TablePrinter) PrintStream(ctx context.Context, w io.Writer, rp output.RowProducer) error {
 	mc, _ := consolesize.GetConsoleSize()
 	if mc <= 0 {
-		mc = tableMaxWidth()
+		mc = tableMaxcols()
 	}
 	if mc < 1 {
 		mc = 1000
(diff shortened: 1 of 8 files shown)


base/printers.go
```diff
@@ -55,17 +55,24 @@ type TablePrinter struct{}
 func (pr *TablePrinter) PrintStream(ctx context.Context, w io.Writer, rp output.RowProducer) error {
 	mc, _ := consolesize.GetConsoleSize()
 	if mc <= 0 {
```
kutluhanmetin, line 57: it cannot be lower than zero that GetConsoleSize returns uint16 for both return values.

internal/table/table.go
```diff
@@ -137,3 +164,35 @@ func withColor(c *color.Color, f func()) {
...
 	color.Unset()
 }
+
+func splitLines(text string, maxWidth int) lines {
+	// just a precaution to not allocate the universe
+	if maxWidth > 65535 {
+		maxWidth = 65535
+	}
+	if maxWidth == 0 {
+		panic("splitLines: maxWidth cannot be 0")
```
kutluhanmetin, line 174: Should we panic?

kutluhanmetin (commented):
Just 2 minor comments

base/printers.go
yuce: > At least we can reduce them to a single if block: if mc == 0
``
	if mc <= 0 {
		mc = tableMaxcols()
	}
	if mc < 1 {
		mc = 1000
	}
```
`TableMaxCols` may also return `0`, so need to check the value of `mc` again.

internal/table/table.go
yuce: `maxWidth` should never be `0` in the normal operation, that would be an exceptional case.

kutluhanmetin (approved):
LGTM

PR #240, opened 2023-05-16
[CLC-170] Viridian Command Updates

kutluhanmetin:
- Remove the `--development` flag.
- Rename `--plan` flag of `viridian create-cluster` to `cluster-type` and do not limit the choice to `SERVERLESS`.(The default is `DEVMODE`).
- Add  `viridian list-cluster-types` command.
- Add `viridian import-config` command which imports the cluster configuration from a Viridian cluster.
- Rename `stop-cluster` to `pause-cluster`.
- Revamp the Viridian errors:
  -  Log the full Viridian error, but do not display JSON in the user error message.
  - Do not display “did you login” except there’s an authentication issue.
- Add `--yes` flag to `viridian delete-custom-class`

### DIFF
--- a/base/commands/viridian/viridian_cluster_create.go
+++ b/base/commands/viridian/viridian_cluster_create.go
@@ -3,6 +3,7 @@ package viridian
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/hazelcast/hazelcast-commandline-client/clc"
--- a/base/commands/viridian/viridian_cluster_create.go
+++ b/base/commands/viridian/viridian_cluster_create.go
@@ -27,7 +27,7 @@ Make sure you login before running this command.
 	cc.SetPositionalArgCount(0, 0)
 	cc.AddStringFlag(propAPIKey, "", "", false, "Viridian API Key")
 	cc.AddStringFlag(flagName, "", "", false, "specify the cluster name; if not given an auto-generated name is used.")
-	cc.AddStringFlag(flagClusterType, "", "", false, "type for the cluster")
+	cc.AddStringFlag(flagClusterType, "", viridian.ClusterTypeServerless, false, "type for the cluster")
 	cc.AddStringFlag(flagHazelcastVersion, "", "", false, "version of the Hazelcast cluster")
 	return nil
 }
--- a/base/commands/viridian/viridian_cluster_create.go
+++ b/base/commands/viridian/viridian_cluster_create.go
@@ -14,6 +14,8 @@ import (
 	"github.com/hazelcast/hazelcast-commandline-client/internal/viridian"
 )
 
+const defaultClusterType = "DEVMODE"
+
 type ClusterCreateCmd struct{}
 
 func (cm ClusterCreateCmd) Init(cc plug.InitContext) error {
@@ -27,7 +29,7 @@ Make sure you login before running this command.
 	cc.SetPositionalArgCount(0, 0)
 	cc.AddStringFlag(propAPIKey, "", "", false, "Viridian API Key")
 	cc.AddStringFlag(flagName, "", "", false, "specify the cluster name; if not given an auto-generated name is used.")
-	cc.AddStringFlag(flagClusterType, "", "", false, "type for the cluster")
+	cc.AddStringFlag(flagClusterType, "", defaultClusterType, false, "type for the cluster")
 	return nil
 }
 
--- a/base/commands/viridian/viridian_cluster_create.go
+++ b/base/commands/viridian/viridian_cluster_create.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 
 	"github.com/hazelcast/hazelcast-commandline-client/clc"
 	"github.com/hazelcast/hazelcast-commandline-client/clc/config"
@@ -28,8 +27,7 @@ Make sure you login before running this command.
 	cc.SetPositionalArgCount(0, 0)
 	cc.AddStringFlag(propAPIKey, "", "", false, "Viridian API Key")
 	cc.AddStringFlag(flagName, "", "", false, "specify the cluster name; if not given an auto-generated name is used.")
-	cc.AddStringFlag(flagPlan, "", "", false, "plan for the cluster, supported values: serverless")
-	cc.AddBoolFlag(flagDevelopment, "", false, false, "start a development cluster")
+	cc.AddStringFlag(flagClusterType, "", "", false, "type for the cluster")
 	return nil
 }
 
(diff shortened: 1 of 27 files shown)


internal/viridian/api.go
```diff
@@ -112,17 +114,32 @@ func (a API) DownloadConfig(ctx context.Context, clusterID string) (path string,
...
+}
+
+func (a API) FindClusterType(ctx context.Context, name string) (ClusterType, error) {
+	clusterTypes, err := a.ListClusterTypes(ctx)
+	if err != nil {
+		return ClusterType{}, err
+	}
+	var found ClusterType
+	for _, ct := range clusterTypes {
+		if strings.ToLower(ct.Name) == strings.ToLower(name) {
```
yuce, line 137: Original name is in upper-case, so it would be better to compare upper case strings.

internal/viridian/api.go
```diff
@@ -112,17 +114,32 @@ func (a API) DownloadConfig(ctx context.Context, clusterID string) (path string,
...
+			found = ct
+			break
```
yuce, line 139: Ho about removing `found` and just `return ct, nil` here?

internal/viridian/cluster.go
```diff
@@ -7,34 +7,49 @@ import (
...
-		ct = ClusterTypeDev
+	clusterTypeID := cType.ID
+	planName := cType.Name
+	if cType.Name == "" {
+		clusterTypes, err := a.ListClusterTypes(ctx)
+		if err != nil {
+			return Cluster{}, err
+		}
+		for _, ct := range clusterTypes {
+			if strings.ToLower(ct.Name) == defaultClusterType {
```
yuce, line 41: Let's compare using upper case.

internal/viridian/cluster.go
```diff
@@ -7,34 +7,49 @@ import (
...
+	if err != nil {
+		return Cluster{}, err
 	}
-	ct := ClusterTypeProd
-	if isDev {
-		ct = ClusterTypeDev
+	clusterTypeID := cType.ID
+	planName := cType.Name
+	if cType.Name == "" {
+		clusterTypes, err := a.ListClusterTypes(ctx)
```
yuce, line 36: Why is ListClusterTypes called twice? One here and one in FindClusterType.

internal/viridian/cluster.go
```diff
@@ -7,34 +7,49 @@ import (
...
+		if err != nil {
+			return Cluster{}, err
+		}
+		for _, ct := range clusterTypes {
+			if strings.ToLower(ct.Name) == defaultClusterType {
+				clusterTypeID = ct.ID
+				planName = ct.Name
+				break
```
yuce, line 44: What happens if the loop ends without setting clusterTypeID and planName?

yuce (changes requested):
Here are some ideas...

internal/viridian/cluster.go
kutluhanmetin: this if block become useless when we set `DEVMODE` as default. So, I am removing it and it won't be called twice.

internal/viridian/cluster.go
kutluhanmetin: this if block become useless when we set DEVMODE as default. So, I am removing it.

base/commands/viridian/viridian_cluster_create.go
```diff
@@ -28,8 +27,7 @@ Make sure you login before running this command.
 	cc.SetPositionalArgCount(0, 0)
 	cc.AddStringFlag(propAPIKey, "", "", false, "Viridian API Key")
 	cc.AddStringFlag(flagName, "", "", false, "specify the cluster name; if not given an auto-generated name is used.")
-	cc.AddStringFlag(flagPlan, "", "", false, "plan for the cluster, supported values: serverless")
-	cc.AddBoolFlag(flagDevelopment, "", false, false, "start a development cluster")
+	cc.AddStringFlag(flagClusterType, "", "", false, "type for the cluster")
```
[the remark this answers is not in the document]
kutluhanmetin: added `--hazelcast-version` parameter, so didn't set it to `DEVMODE` by defauilt.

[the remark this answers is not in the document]
yuce: Could you set the default to `SERVERLESS` ?

base/commands/viridian/viridian_cluster_types.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+	verbose := ec.Props().GetBool(clc.PropertyVerbose)
+	csi, stop, err := ec.ExecuteBlocking(ctx, func(ctx context.Context, sp clc.Spinner) (any, error) {
+		sp.SetText("Retrieving cluster types")
+		cs, err := api.ListClusterTypes(ctx)
+		if err != nil {
+			return nil, err
+		}
+		return cs, nil
```
yuce, line 41: How about just `return pi.ListClusterTypes(ctx)` ?

docs/modules/ROOT/pages/clc-viridian.adoc
```diff
@@ -83,23 +111,25 @@ Parameters:
...
 
-|`--plan`
+|`--cluster-type`
 |Optional
-|Sets the plan for the created cluster.
-|serverless
+|Sets the cluster-type for the created cluster.
 
-|`--development`
+WARNING: You can set `DEVMODE` to `SERVERLESS` without giving the `--hazelcast-version`. However, if you set `--cluster-type` to `DEVMODE` without providing `--hazelcast-version`, it will be automatically converted to `SERVERLESS`.
```
yuce, line 126: I think we can remove this warning, it's an implementation detail.

docs/modules/ROOT/pages/managing-viridian-clusters.adoc
```diff
@@ -31,12 +31,14 @@ To create a new cluster on Viridian, you can use the following command:
 
 [source, bash]
 ----
-clc viridian create-cluster --name my-cluster --development
+clc viridian create-cluster --name my-cluster --cluster-type DEVMODE --hazelcast-version 5.3.0-BETA-2
```
yuce, line 34:
```suggestion
clc viridian create-cluster --name my-cluster
```

docs/modules/ROOT/pages/managing-viridian-clusters.adoc
```diff
@@ -31,12 +31,14 @@ To create a new cluster on Viridian, you can use the following command:
...
 Imported configuration: my-cluster
 OK
 ----
 This command creates a development cluster named `my-cluster`.
 
+WARNING: You can set `DEVMODE` to `SERVERLESS` without giving the `--hazelcast-version`. However, if you set `--cluster-type` to `DEVMODE` without providing `--hazelcast-version`, it will be automatically converted to `SERVERLESS`.
```
yuce, line 40: Let's remove this warning, it's an implementation detail.

errors/error.go
```diff
@@ -43,3 +45,41 @@ type HTTPError interface {
...
+}
+
+func NewHTTPClientError(code int, body []byte) error {
+	err := HTTPClientError{
+		code:    code,
+		rawResp: string(body),
+		// it can be overwritten
+		text: "an unexpected error occurred, please check logs for details",
+	}
+	type ErrResp struct {
```
yuce, line 62:
```suggestion
	type ErrorResponse struct {
```

Also, could you move this type out of the function?

errors/error.go
```diff
@@ -43,3 +45,41 @@ type HTTPError interface {
...
 	Code() int
 }
+
+type HTTPClientError struct {
+	code    int
+	text    string
+	rawResp string
+}
+
+func NewHTTPClientError(code int, body []byte) error {
```
yuce, line 55: Would moving `HTTPClientError` and `NewHTTPClientError` to `internal/viridian/errors.go`  make sense?

internal/viridian/api.go
```diff
@@ -112,17 +114,30 @@ func (a API) DownloadConfig(ctx context.Context, clusterID string) (path string,
...
 		if c.ID == idOrName || c.Name == idOrName {
-			return c.ID, nil
+			return c, nil
+		}
+	}
+	return Cluster{}, fmt.Errorf("no such cluster found: %s", idOrName)
+}
+
+func (a API) FindClusterType(ctx context.Context, name string) (ClusterType, error) {
+	clusterTypes, err := a.ListClusterTypes(ctx)
```
yuce, line 131: I would rename `clusterTypes` to `cts`

yuce (changes requested):
A few more ideas

yuce (approved):
Looks good!

PR #245, opened 2023-05-18
Various Updates

yuce:
- Use pause consistently
- Updated doc for MacOS
- Do not print an error message on user cancelled

### DIFF
--- a/clc/cmd/exec_context.go
+++ b/clc/cmd/exec_context.go
@@ -234,27 +234,26 @@ func (ec *ExecContext) WrapResult(f func() error) error {
 	took := time.Since(t)
 	verbose := ec.Props().GetBool(clc.PropertyVerbose)
 	if err != nil {
-		if errors.Is(err, context.Canceled) {
-			I2(fmt.Fprintln(ec.stderr, "User cancelled"))
+		if errors.Is(err, context.Canceled) || errors.Is(err, cmderrors.ErrUserCancelled) {
+			return nil
+		}
+		var errStr string
+		var msg string
+		var httpErr cmderrors.HTTPError
+		if errors.As(err, &httpErr) {
+			errStr = makeErrorStringFromHTTPResponse(httpErr.Text())
 		} else {
-			var errStr string
-			var msg string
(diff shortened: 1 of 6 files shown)


docs/modules/ROOT/pages/install-clc.adoc
```diff
@@ -30,6 +36,12 @@ ZIP (Apple Silicon)::
 +
 . Go to the https://github.com/hazelcast/hazelcast-commandline-client/releases/latest[latest release page], and locate the ARM compressed file (`hazelcast-clc_v{full-version}_darwin_arm64.zip`).
 . Download and unzip the file.
+. Remove the `clc` binary from quarantine if you get a security warning when running it:
++
+[source,shell,subs="attributes"]
+----
+xattr -d com.apple.quarantine hazelcast-clc_v{full-version}_darwin_amd64/clc
```
kutluhanmetin, line 43: I think this should be arm, right?

clc/cmd/exec_context.go
```diff
@@ -234,27 +234,26 @@ func (ec *ExecContext) WrapResult(f func() error) error {
...
-				msg = fmt.Sprintf("Error: %s", errStr)
-			}
-			if ec.Interactive() {
-				I2(fmt.Fprintln(ec.stderr, color.RedString(msg)))
-			} else {
-				I2(fmt.Fprintln(ec.stderr, msg))
-			}
+			errStr = err.Error()
+		}
+		if verbose {
```
kutluhanmetin, line 248: I think this is optional: we can get rid of else statements by initializing them with the values inside `else`s.

yuce: I am also not a fan of the `else`s, but in this case that would require calling functions more than once.

docs/modules/ROOT/pages/install-clc.adoc
yuce: Thanks, updated at: https://github.com/hazelcast/hazelcast-commandline-client/pull/245/commits/8aa04515f273aeca9661a01a90277b98ba1fcc0f

PR #250, opened 2023-05-23
[CLC-173] stream-logs Command

yuce:
This PR implements the `viridian stream-logs` command. The client is disconnected to the server, since the server closes the socket after 1 minute. In order to overcome that problem, the client connects again and discards the previously read logs.


docs/modules/ROOT/pages/clc-viridian.adoc
```diff
@@ -296,11 +293,49 @@ Parameters:
...
+Parameters:
+
+[cols="1m,1a,2a,1a"]
+|===
+|Parameter|Required|Description|Default
+
+|`--api-key`
+|Optional
+|Sets the API key. Overrides the `CLC_VIRIDIAN_API_KEY` environment variable. If not given, one of the existing API keys will be used.
+|
```
mtyazici, line 325: Can we also add `cluster-ID/name` as a parameter?

yuce: It seems we didn't add that to other `viridian` commands.

mtyazici (approved):
LGTM

PR #251, opened 2023-05-23
[CLC-188] Refactor Jet Support

yuce:
* We will need to use Jet functionality outside of `jet` command, so moved core Jet code to a separate package.
* Added support for submitting the job from a reader.

### DIFF
--- a/base/commands/job/job_submit.go
+++ b/base/commands/job/job_submit.go
@@ -1,36 +1,26 @@
 package job
 
 import (
-	"bytes"
 	"context"
-	"crypto/sha256"
 	"fmt"
-	"io"
 	"math"
-	"math/rand"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/hazelcast/hazelcast-go-client"
-	"github.com/hazelcast/hazelcast-go-client/cluster"
-	"github.com/hazelcast/hazelcast-go-client/types"
 
 	"github.com/hazelcast/hazelcast-commandline-client/clc"
 	"github.com/hazelcast/hazelcast-commandline-client/clc/cmd"
 	"github.com/hazelcast/hazelcast-commandline-client/clc/paths"
 	. "github.com/hazelcast/hazelcast-commandline-client/internal/check"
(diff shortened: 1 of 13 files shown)


internal/jet/job.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/hazelcast/hazelcast-go-client"
+	"github.com/hazelcast/hazelcast-go-client/types"
+
+	"github.com/hazelcast/hazelcast-commandline-client/internal/cluster"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/log"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/proto/codec"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/proto/codec/control"
+)
+
+type spinner interface {
```
kutluhanmetin, line 18: Do you remember that you suggested to pass an updater function while I was implementing custom class download/upload (`progressSetter`). Do we really need this extra interface as we already have a spinner interface inside `clc/common.go`?

base/commands/job/job_submit.go
```diff
@@ -1,36 +1,26 @@
...
 	"github.com/hazelcast/hazelcast-commandline-client/internal/plug"
-	"github.com/hazelcast/hazelcast-commandline-client/internal/proto/codec"
 )
 
 const (
-	// see: https://github.com/hazelcast/hazelcast/issues/24285
-	envExperimentalCalculateHashWorkaround = "CLC_EXPERIMENTAL_WORKAROUND_24285"
-	minServerVersion                       = "5.3.0-BETA-2"
-	defaultBatchSize                       = 2 * 1024 * 1024 // 10MB
+	minServerVersion = "5.3.0"
```
kutluhanmetin, line 23: it is not directly related to this line, but maybe you would like to use `github.com/hashicorp/go-version` package to compare versions instead of `internal/version.go`.

internal/jet/job.go
yuce: Yes. The primary motivation for my comment was decoupling code in the API from the commands. Doing that via n interface may be better, since later we may have to add more methods to this interface.

kutluhanmetin (approved):

PR #252, opened 2023-05-24
DOCS-521_Update MC and CLC versions

angelasimms:
Updated antora.yml:

- Links to Management Center showing as broken as using page-latest-supported-mc: '5.3-snapshot' rather than '5.3.1-snapshot'
- Added new snapshot details.

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -2,21 +2,21 @@ name: clc
 title: Hazelcast CLC
 start_page: overview.adoc
 # Version in the URL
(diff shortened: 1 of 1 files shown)


yuce:
@angelasimms Thanks for this PR. We'll release 5.3.1 next, and there's quite a bit of time until we release 5.4.0, so I think it would be better to make it 5.3.1-snapshot.

Serdaro (approved):

Serdaro (approved):

PR #254, opened 2023-05-25
DOCS-504 Updates to Jet Job Tutorial

angelasimms:
Some updates to tutorial, including:

- Changes to step 4 to make it easier to use the example Jet job
- Added steps to stop looping statements

**Issues/errors**
- `clc -c dev job submit build/libs/simple-streaming-pipeline-1.0-SNAPSHOT-all.jar –wait` gave the following error `Error: --wait requires the --name to be set`. When I removed `--wait`, the command ran. For now, I have left the command as is.
- `while true; do clc -c dev map size -n my-map; sleep 3; done` does not run for Windows users. I created a small batch file and ran it. Do you want to provide this alternative for Windows users?

```
@echo off
:start
clc -c dev map size -n my-map
timeout 5
goto start

```
- `clc -c dev job cancel prime-job` gave the following error `Error: invalid job ID`.  As a workaround, I suspended the job through management center.

I will submit a separate PR to main with the same changes, when this one is merged.

### DIFF
--- a/docs/modules/ROOT/pages/jet-job-management.adoc
+++ b/docs/modules/ROOT/pages/jet-job-management.adoc
@@ -96,7 +96,7 @@ Submitting the sample Jet job generates a stream of numbers and creates a link:h
 +
 [source,shell]
 ----
-clc -c dev job submit build/libs/simple-streaming-pipeline-1.0-SNAPSHOT-all.jar --wait
+clc -c dev job submit build/libs/simple-streaming-pipeline-1.0-SNAPSHOT-all.jar --name simple-pipeline --wait
 ----
 +
 The Hazelcast CLC waits until the job starts to run.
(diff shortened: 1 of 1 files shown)


yuce:
Hi Angela,

> clc -c dev job submit build/libs/simple-streaming-pipeline-1.0-SNAPSHOT-all.jar –wait gave the following error Error: --wait requires the --name to be set. When I removed --wait, the command ran. For now, I have left the command as is.
The pipeline has a name, but that's in its code so CLC doesn't know about it. Would it be possible for you to add the following parameter to the problematic line? `--name simple-pipeline` ?

> I created a small batch file and ran it. Do you want to provide this alternative for Windows users?
Thanks, that's a great idea. I think it would be better to convert the one-liner for Unix to a shell script too, so we can show them as tabs.

> clc -c dev job cancel prime-job gave the following error Error: invalid job ID.
I've changed Muhammet's example a bit, but forgot to change the job name. Could you change the line to: `clc -c dev job cancel simple-pipeline` ?

yuce (approved):
Looks good!

yuce (approved):
Thanks, looks great!

Serdaro (approved):
LVGTM

PR #255, opened 2023-05-26
[CLC-187]: Add Global `--timeout` Flag

kutluhanmetin:
The duration is a string in the form of `AMOUNTunit` where `AMOUNT` is an `integer`, and `unit` is one of the following:

- `ms: milliseconds`

- `s: seconds`

- `m: minutes`

- `h: hours`

Underscore (_) character is ignored in the `AMOUNT`. `unit` is optional and defaults to `ms` if not specified.

The following are a few of the valid DURATIONs:

- `5: 5 seconds`

- `5s: 5 seconds`

- `10000ms: 10000 milliseconds`

- `10_000_000ms: 10000000 milliseconds`


yuce (changes requested):

kutluhanmetin:
Tested with `clc -c c01 sql 'select * from table(generate_stream(1))'`

yuce (approved):
Looks good!

PR #261, opened 2023-05-29
[CLC-195]: Fix Auth Error Message

kutluhanmetin:


base/commands/viridian/common.go
```diff
@@ -108,6 +109,9 @@ func handleErrorResponse(ec plug.ExecContext, err error) error {
 	// if it is a http client error, return the simplified error to user
 	var err2 viridian.HTTPClientError
 	if errors.As(err, &err2) {
+		if err2.Code() == http.StatusUnauthorized {
```
yuce, line 112: I think it would be better to rename `err2` to something like `clientErr`, or `ce` to make it shorter.

yuce (changes requested):
Just a minor comment

yuce (approved):

PR #263, opened 2023-05-30
Add map load-all command [CLC-178]

mtyazici:


base/commands/map/map_load_all.go
```diff
@@ -0,0 +1,64 @@
...
+	return nil
+}
+
+func (mc *MapLoadAllCommand) Exec(ctx context.Context, ec plug.ExecContext) error {
+	mapName := ec.Props().GetString(mapFlagName)
+	ci, err := ec.ClientInternal(ctx)
+	if err != nil {
+		return err
+	}
+	keys := []hazelcast.Data{}
```
yuce, line 36:
```suggestion
	var keys []hazelcast.Data
```

internal/it/map_tester.go
```diff
@@ -9,18 +9,25 @@ import (
 	"github.com/hazelcast/hazelcast-commandline-client/internal/check"
 )
 
-func WithMap(tcx TestContext, fn func(m *hz.Map)) {
-	name := NewUniqueObjectName("map")
+func WithMap(tcx TestContext, mn string, fn func(m *hz.Map)) {
```
yuce, line 12: How about using a different name for this function instead of renaming the other. Maybe `WithNamedMap` ?

yuce (changes requested):
A few ideas...

yuce (approved):
Looks good!

PR #269, opened 2023-06-02
Reorganize viridian get-cluster command

yuce:
Reorganized the columns in the result of `viridian get-cluster`.


yuce:
Trivial, merging it myself

PR #281, opened 2023-06-07
[CLC-190] Add MultiMap Command

kutluhanmetin:


base/commands/multimap/multimap_lock.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+	keyStr := ec.Args()[0]
+	keyData, err := makeKeyData(ec, ci, keyStr)
+	if err != nil {
+		return err
+	}
+	req := codec.EncodeMultiMapLockRequest(mmName, keyData, 0, ttl, 0)
+	_, stop, err := ec.ExecuteBlocking(ctx, func(ctx context.Context, sp clc.Spinner) (any, error) {
+		sp.SetText(fmt.Sprintf("Locking multimap %s", mmName))
```
mtyazici, line 41:
```suggestion
		sp.SetText(fmt.Sprintf("Locking key of multimap %s", mmName))
```

Maybe we can even print the key as well

base/commands/multimap/multimap_unlock.go
```diff
@@ -0,0 +1,55 @@
...
+		return ci.InvokeOnKey(ctx, req, keyData, nil)
+	})
+	if err != nil {
+		return err
+	}
+	stop()
+	if err != nil {
+		return err
+	}
+	stop()
```
mtyazici, line 49:
```suggestion
```

base/commands/multimap/multimap_unlock.go
```diff
@@ -0,0 +1,55 @@
...
+		return err
+	}
+	keyStr := ec.Args()[0]
+	keyData, err := makeKeyData(ec, ci, keyStr)
+	if err != nil {
+		return err
+	}
+	req := codec.EncodeMultiMapUnlockRequest(mmName, keyData, 0, 0)
+	_, stop, err := ec.ExecuteBlocking(ctx, func(ctx context.Context, sp clc.Spinner) (any, error) {
+		sp.SetText(fmt.Sprintf("Unlocking multimap %s", mmName))
```
mtyazici, line 39:
```suggestion
		sp.SetText(fmt.Sprintf("Unlocking key of multimap %s", mmName))
```

base/commands/multimap/multimap_lock.go
```diff
@@ -0,0 +1,53 @@
...
+	ci, err := ec.ClientInternal(ctx)
+	if err != nil {
+		return err
+	}
+	keyStr := ec.Args()[0]
+	keyData, err := makeKeyData(ec, ci, keyStr)
+	if err != nil {
+		return err
+	}
+	req := codec.EncodeMultiMapLockRequest(mmName, keyData, 0, ttl, 0)
```
mtyazici, line 39: Do we need to expose the TTL for lock operations? Go client implementation does not support ttl for lock operations. Maybe we could just use client implementation since we are not encoding/decoding any data?

mtyazici (changes requested):
I left some comments and questions.

base/commands/multimap/multimap_lock.go
mtyazici: I saw that client has `lockWithLease` which supports the TTL but it also has support for the lock context. So it might be better to just use that function.

mtyazici (dismissed):

yuce (approved):
Looks good!

PR #290, opened 2023-06-27
Refactored job commands to use coordinator for SQL jobs

yuce:


base/commands/job/common.go
```diff
@@ -60,89 +63,91 @@ func idToString(id int64) string {
...
+			coord = conns[0].MemberUUID()
+		}
+		return jid, j.TerminateJob(ctx, jid, terminateMode, coord)
 	})
 	if err != nil {
 		return err
 	}
 	stop()
+	verbose := ec.Props().GetBool(clc.PropertyVerbose)
+	err = nil
```
kutluhanmetin, line 108: I think this is a leftover

kutluhanmetin, line 108: We already returned the error if it is not nil, we don't need to set it to nil in here.

base/commands/job/common.go
```diff
@@ -60,89 +63,91 @@ func idToString(id int64) string {
...
+			}
+			coord = conns[0].MemberUUID()
+		}
+		return jid, j.TerminateJob(ctx, jid, terminateMode, coord)
 	})
 	if err != nil {
 		return err
 	}
 	stop()
+	verbose := ec.Props().GetBool(clc.PropertyVerbose)
```
kutluhanmetin, line 107: I think we don't need to initialize verbose, we can directly use `ec.Props().GetBool(clc.PropertyVerbose)`. I think it is more readable. Same for `wait := ec.Props().GetBool(flagWait)` as well.

base/commands/job/common.go
yuce: Addressed at: https://github.com/hazelcast/hazelcast-commandline-client/pull/290/commits/d05e883adb10b872c1f6cd9577db28dfcb3bddce

base/commands/job/common.go
```diff
@@ -60,89 +63,91 @@ func idToString(id int64) string {
...
-			return err
+	}
+	if err != nil {
+		if verbose {
+			ec.PrintlnUnnecessary(fmt.Sprintf("Job %sed: %s", name, idToString(jidv.(int64))))
 		}
+		return nil
 	}
-	return nil
+	return err
```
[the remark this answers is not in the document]
yuce: Addressed at: https://github.com/hazelcast/hazelcast-commandline-client/pull/290/commits/d05e883adb10b872c1f6cd9577db28dfcb3bddce

base/commands/job/common.go
yuce: Addressed at https://github.com/hazelcast/hazelcast-commandline-client/pull/290/commits/d05e883adb10b872c1f6cd9577db28dfcb3bddce and https://github.com/hazelcast/hazelcast-commandline-client/pull/290/commits/1d87b6c2ace40bb051dadb098ffc91ce68fcf73e

kutluhanmetin (approved):
LGTM!

PR #292, opened 2023-07-05
[CLC-19]: `\di` Command to List Indexes

kutluhanmetin:


base/commands/shell_script_common.go
```diff
@@ -48,7 +47,7 @@ func makeEndLineFunc() shell.EndLineFn {
 	}
 }
 
-func makeTextFunc(m *cmd.Main, ec plug.ExecContext, verbose, ignoreErrors, echo bool, sf shortcutFunc) shell.TextFn {
+func makeTextFunc(m *cmd.Main, ec plug.ExecContext, verbose bool, sf shortcutFunc) shell.TextFn {
```
yuce, line 50: Let's keep the signature as is even though `ie` and `echo` are not used at the moment.

clc/shell/common.go
```diff
@@ -1,62 +1,98 @@
...
 	}
-	return stmt, nil
+	return f, nil
 }
 
 func InteractiveHelp() string {
 	return `
 Shortcut Commands:
+	\di			  List Indexes
+	\di	 MAPPING  List Indexes for a specific mapping
```
yuce, line 95: Could you align these lines?
So it looks like:
```
\di           List Indexes
\di  MAPPING  List Indexes for a specific mapping
\dm           List mappings
\dm  MAPPING  Display information about a mapping
\dm+ MAPPING  Describe a mapping
```

yuce (changes requested):
Some ideas...

yuce (approved):
Looks good!

PR #295, opened 2023-07-17
[CLC-165]: Add details to Snapshot List Command Output

kutluhanmetin:


internal/serialization/snapshot.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return nil
+}
+
+func (SnapshotFactory) FactoryID() int32 {
+	return snapshotFactoryID
+}
+
+func (s *Snapshot) WriteData(output serialization.DataOutput) {
+
```
yuce, line 42:
```suggestion
// not used
```

internal/serialization/snapshot.go
```diff
@@ -0,0 +1,53 @@
...
+	return snapshotClassID
+}
+
+type SnapshotFactory struct{}
+
+func (SnapshotFactory) Create(classID int32) serialization.IdentifiedDataSerializable {
+	if classID == snapshotClassID {
+		return &Snapshot{}
+	}
+	return nil
```
yuce, line 34: It is better to panic here instead of returning `nil` so the actual error is known.

yuce (changes requested):
Two minor ideas...

internal/serialization/snapshot.go
kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/295/commits/ea6f747894efa10efd50b5296f03673e644117cb

internal/serialization/snapshot.go
kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/295/commits/ea6f747894efa10efd50b5296f03673e644117cb

yuce (approved):
Looks good!

PR #297, opened 2023-07-18
Stage support

yuce:
* Colored output for `OK` and `Fail` status,
* Added the Stage component for multi-step status output.


clc/cmd/exec_context.go
```diff
@@ -314,13 +325,18 @@ func (s *simpleSpinner) Start() {
...
+}
+
 func (s *simpleSpinner) SetText(text string) {
 	s.text = text
 	if text == "" {
 		s.sp.Prefix("")
 		return
 	}
-	s.sp.Prefix(text + cancelMsg)
+	s.sp.Prefix("      " + text + cancelMsg)
```
kutluhanmetin, line 339: Shall we use strings.Repeat here to make it more readable?

kutluhanmetin (approved):

clc/cmd/exec_context.go
yuce: I've added that, but it made the code less clean.
I'll keep it if it's OK.

PR #309, opened 2023-07-27
Add Badger as local store [CLC-234]

mtyazici:


clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (l badgerLogger) Errorf(log string, args ...any) {
+	l.logger.Debugf(log, args...)
+}
+func (l badgerLogger) Warningf(log string, args ...any) {
+	l.logger.Debugf(log, args...)
+}
+func (l badgerLogger) Infof(log string, args ...any) {
+	l.logger.Debugf(log, args...)
```
yuce, line 72: Does it make sense to add a prefix, like `Store ERROR: `, etc. to these log lines so it's possible to distinguish between them?

clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+func (s *Store) GetKeysWithPrefix(prefix string) ([][]byte, error) {
+	keys := [][]byte{}
+	pb := []byte(prefix)
+	opts := badger.DefaultIteratorOptions
+	opts.PrefetchValues = false
+	err := s.db.View(func(txn *badger.Txn) error {
+		it := txn.NewIterator(opts)
+		defer it.Close()
+		for it.Seek(pb); it.ValidForPrefix(pb); it.Next() {
+			keyCopy := it.Item().KeyCopy(nil)
```
yuce, line 118:
```suggestion
			c := it.Item().KeyCopy(nil)
```

clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return keys, nil
+}
+
+type updateFn = func(current []byte, found bool) []byte
```
yuce, line 129:
```suggestion
type updateFunc = func(current []byte, found bool) []byte
```

clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (s *Store) UpdateEntry(key []byte, f updateFn) error {
```
yuce, line 131: Since `UpdateEntry` is exported, `updateFn` should also be exported.

clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+	var item []byte
+	err := s.db.Update(func(txn *badger.Txn) error {
+		it, err := txn.Get(key)
+		found := true
+		if err != nil {
+			if errors.Is(err, badger.ErrKeyNotFound) {
+				found = false
+			} else {
+				return err
+			}
```
yuce, line 141:
```suggestion
			if !errors.Is(err, badger.ErrKeyNotFound) {
				return err
			}
			found = false
```

clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		newVal := f(item, found)
+		return txn.Set(key, newVal)
+	})
+	return err
+}
+
+type foreachFn = func(key, val []byte)
```
yuce, line 154:
```suggestion
type foreachFunc = func(key, val []byte)
```

clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (s *Store) RunForeachWithPrefix(prefix string, f foreachFn) error {
```
yuce, line 156: Since `RunForeachWithPrefix` is exporeted, `foreachFn` should also be exported.

clc/store/store.go
yuce: How about:
```
type ForeachFunc = func(key, val []byte) (ok bool, err error)
```
So if the functio returns `ok == false` we can exit the loop early.
And if there's an error we can return it from `RunForeachWithPrefix`.

clc/store/store.go
```diff
@@ -0,0 +1,189 @@
...
+		it := txn.NewIterator(badger.DefaultIteratorOptions)
+		defer it.Close()
+		for it.Seek(p); it.ValidForPrefix(p); it.Next() {
+			item := it.Item()
+			k := item.KeyCopy(nil)
+			v, err := item.ValueCopy(nil)
+			if err != nil {
+				return err
+			}
+			f(k, v)
```
yuce, line 168: Would reusing `k` after the first call to `item.KeyCopy` to optimize creating byte arrays make sense? Same for `v`.

clc/store/store_test.go
```diff
@@ -0,0 +1,162 @@
...
+	badger "github.com/dgraph-io/badger/v4"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hazelcast/hazelcast-commandline-client/internal/check"
+	"github.com/hazelcast/hazelcast-commandline-client/internal/log"
+)
+
+func bytes(s string) []byte {
+	return []byte(s)
+}
```
yuce, line 17: I think it's better to remove this function, since `[]byte(s)` is well known and it's not much longer than `bytes(s)`.

clc/store/store_test.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func WithTempDir(fn func(string)) {
```
yuce, line 142: Might be nice to add the parameter name to the callback for documentation.
```suggestion
func WithTempDir(fn func(dir string)) {
```

yuce (changes requested):
Some ideas...

clc/store/store.go
mtyazici: Alright, I will change the signature

clc/store/store.go
mtyazici: Could you explain it more? I am not sure if I got it correctly.

yuce: It's OK to create the byte array for each entry for now.
We'll reconsider if we have any performance problems with this code.

yuce (approved):
Looks good!

PR #312, opened 2023-07-28
Use Python code sample instead of Go

yuce:
- Use Python code sample instead of Go
- Fixed a couple of issues

### DIFF
--- a/base/commands/config/config_import.go
+++ b/base/commands/config/config_import.go
@@ -23,16 +23,16 @@ Currently importing only Viridian connection configuration is supported.
 	
 1. On Viridian console, visit:
 	
-	Dashboard -> Connect Client -> Quick connection guide -> Go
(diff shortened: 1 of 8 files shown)


base/commands/config/config_import.go
```diff
@@ -23,16 +23,16 @@ Currently importing only Viridian connection configuration is supported.
 	
 1. On Viridian console, visit:
 	
-	Dashboard -> Connect Client -> Quick connection guide -> Go
```
kutluhanmetin, line 26: We should also update the steps in documentation file `clc-config.adoc`

yuce: Good catch, updated at: https://github.com/hazelcast/hazelcast-commandline-client/pull/312/commits/f226f9ee82e278e03905c159df92ad9e88f8ed45

kutluhanmetin (approved):

PR #318, opened 2023-08-02
Add config name into CLC command promt [CLC-242]

mtyazici:


base/commands/shell.go
```diff
@@ -103,6 +105,33 @@ func (cm *ShellCommand) ExecInteractive(ctx context.Context, ec plug.ExecContext
 	return sh.Start(ctx)
 }
 
+func prompt(cfgPath string) string {
```
yuce, line 108:
```suggestion
func makePrompt(cfgPath string) string {
```

base/commands/shell.go
```diff
@@ -103,6 +105,33 @@ func (cm *ShellCommand) ExecInteractive(ctx context.Context, ec plug.ExecContext
...
+	}
+	// Best effort for absolute path
+	if !filepath.IsAbs(cfgPath) {
+		p, err := filepath.Abs(cfgPath)
+		if err == nil {
+			cfgPath = p
+		}
+	}
+	pd := parentDir(cfgPath)
+	return fmt.Sprintf("CLC (%s) > ", dotify(pd, 12))
```
yuce, line 120:
```suggestion
	return fmt.Sprintf("%s> ", dotify(pd, 12))
```

base/commands/shell.go
```diff
@@ -103,6 +105,33 @@ func (cm *ShellCommand) ExecInteractive(ctx context.Context, ec plug.ExecContext
 	return sh.Start(ctx)
 }
 
+func prompt(cfgPath string) string {
+	if cfgPath == "" {
+		return "CLC> "
```
yuce, line 110: `cfgPath` should never be a blank string, but just in case it is...
```suggestion
		return "> "
```

base/commands/shell.go
```diff
@@ -103,6 +105,33 @@ func (cm *ShellCommand) ExecInteractive(ctx context.Context, ec plug.ExecContext
...
+	}
+	// Best effort for absolute path
+	if !filepath.IsAbs(cfgPath) {
```
yuce, line 113: Is this check necessary?

base/commands/shell.go
```diff
@@ -103,6 +105,33 @@ func (cm *ShellCommand) ExecInteractive(ctx context.Context, ec plug.ExecContext
...
+		}
+	}
+	pd := parentDir(cfgPath)
+	return fmt.Sprintf("CLC (%s) > ", dotify(pd, 12))
+}
+
+func parentDir(path string) string {
+	dirs := filepath.Dir(path)
+	return filepath.Base(dirs)
+}
```
yuce, line 126: Could you move this function to `paths.ParentDir` ?

base/commands/shell.go
```diff
@@ -103,6 +105,33 @@ func (cm *ShellCommand) ExecInteractive(ctx context.Context, ec plug.ExecContext
...
+
+func dotify(s string, l int) string {
+	if len(s) < 3 || len(s) < l {
+		return s
+	}
+	return fmt.Sprintf("%s...", s[:l])
+}
+
```
yuce, line 134: Could you move this function to `str.MaybeShorten` ?

yuce (changes requested):
A few ideas...

base/commands/shell.go
mtyazici: Sure 👍🏼

base/commands/shell.go
mtyazici: It's not necessary, I will take the statemens inside to the root level

yuce (dismissed):
Looks good!

yuce (approved):
Looks good!

PR #320, opened 2023-08-03
Updated project creator

yuce:
* Only lowercase lettes, numbers and the underscore character are allowed as placeholders
* Converts properties to valid placeholders by replacing `.` with `_`.
* Adds the optional `--output-dir` flag which defaults to the template name.
* Updated project creator test, so it is self sufficient. The downside is, git behaviour is not tested. Need to introduce another test later.


base/commands/project/utils.go
```diff
@@ -31,22 +31,13 @@ func loadFromDefaults(templateDir string) (map[string]string, error) {
 	if err = parseYAML("", b, props); err != nil {
 		return nil, err
 	}
-	props = camelizeMapKeys(props)
-	if err != nil {
-		return nil, err
+	if key, ok := validateMap(props); !ok {
```
kutluhanmetin, line 34: `validateMap` is so generic. I would suggest something like `validateKeyFormat`, `validatePropKeys` etc.

docs/modules/ROOT/pages/clc-project.adoc
```diff
@@ -42,9 +42,8 @@ Parameters:
 
 |`placeholder-values`
 |Optional
-|Template placeholder values can be specified as key-value pairs. You can use letters and numbers in keys. Example: `key`=`value`
+|Template placeholder values can be specified as key-value pairs. You can use lowercase letters, numbers and the underscore character in keys. Example: `key`=`value`
```
kutluhanmetin, line 45:
```suggestion
|Template placeholder values can be specified as key-value pairs. You can use lowercase letters, numbers and the underscore character in keys. Example: `my_key1`=`value`
```

yuce: Addressed at: https://github.com/hazelcast/hazelcast-commandline-client/pull/320/commits/f74c77102dd332b8f41fdbeffe27adf0b5333820

kutluhanmetin (approved):
looks good

PR #326, opened 2023-08-04
[CLC-245]: Add Command to List Project Templates

kutluhanmetin:


base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	for _, t := range ts {
+		templates = append(templates, Template{Name: t, Source: "local"})
+	}
+	return templates, nil
+}
+
+func fetchTemplates() ([]Template, error) {
+	var templates []Template
+	resp, err := http.Get(fetchURL())
```
yuce, line 125: Should the results be paged? There maybe hundreds of repos in the org.

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		return []Template{}, err
+	}
+	respData, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return []Template{}, err
```
yuce, line 131:
```suggestion
		return nil, err
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+		templates = append(templates, Template{Name: t, Source: "local"})
+	}
+	return templates, nil
+}
+
+func fetchTemplates() ([]Template, error) {
+	var templates []Template
+	resp, err := http.Get(fetchURL())
+	if err != nil {
+		return []Template{}, err
```
yuce, line 127:
```suggestion
		return nil, err
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	respData, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return []Template{}, err
+	}
+	var data []map[string]any
+	err = json.Unmarshal(respData, &data)
+	if err != nil {
+		return []Template{}, err
```
yuce, line 136:
```suggestion
		return nil, err
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	for _, d := range data {
+		var tName string
+		var ok bool
+		if tName, ok = d["full_name"].(string); !ok {
+			return []Template{}, errors.New("error fetching repositories in the organization")
```
yuce, line 142:
```suggestion
			return nil, errors.New("error fetching repositories in the organization")
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+		name := sName[1]
+		templates = append(templates, Template{Name: name, Source: source})
+	}
+	return templates, nil
+}
+
+func updateNextFetchTime(sa *store.StoreAccessor) error {
+	_, err := sa.WithLock(func(s *store.Store) (any, error) {
+		return nil, s.SetEntry(bytes(nextFetchTimeKey),
+			bytes(strconv.FormatInt(time.Now().Add(cacheRefreshInterval).Unix(), 10)))
```
yuce, line 155: Would it make sense to move the value to a variable?

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	})
+	return err
+}
+
+func fetchURL() string {
```
yuce, line 160:
```suggestion
func makeRepositoriesURL() string {
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	s := strings.TrimPrefix(templateOrgURL(), "https://github.com/")
+	ss := strings.ReplaceAll(s, "/", "")
+	return fmt.Sprintf("https://api.github.com/users/%s/repos", ss)
+}
+
+func isFetch(s *store.StoreAccessor) (bool, error) {
```
yuce, line 166:
```suggestion
func canFetchTemplateList(s *store.StoreAccessor) (bool, error) {
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	stop()
+	return ec.AddOutputRows(ctx, rows...)
+}
+
+func listTemplates(logger log.Logger, isLocal bool, isForce bool) ([]Template, error) {
+	sa := store.NewStoreAccessor(paths.Join(paths.Home(), storeFolder), logger)
+	if isLocal {
+		return listLocalTemplates()
+	}
+	if fetch, err := isFetch(sa); err != nil {
```
yuce, line 94: Can we continue and list the templates if `isFetch` returns an error?
Logging the error in DEBUG would be good.

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+		return []Template{}, err
+	} else if fetch || isForce {
+		ts, err := fetchTemplates()
+		if err != nil {
+			return nil, err
+		}
+		err = updateCache(sa, ts)
```
yuce, line 101: Let's log the error as DEBUG and continue listing the templates.

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+func updateCache(sa *store.StoreAccessor, templates []Template) error {
+	b, err := json.Marshal(templates)
+	if err != nil {
+		return err
+	}
+	_, err = sa.WithLock(func(s *store.Store) (any, error) {
+		err = s.DeleteEntriesWithPrefix(templatesKey)
+		if err != nil {
+			return nil, err
+		}
```
yuce, line 197: Do we have to delete the template key?
Wouldn't setting the new value overwrite the older one?

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+		err = s.SetEntry(bytes(templatesKey), b)
+		if err != nil {
+			return nil, err
+		}
+		return nil, nil
+	})
+	if err != nil {
+		return err
+	}
+	if err = updateNextFetchTime(sa); err != nil {
```
yuce, line 207: Would it make sense to update the fetch time in the .SetEntry block above?

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	return nil
+}
+
+func listFromCache(sa *store.StoreAccessor) ([]Template, error) {
+	var templates []Template
+	b, err := sa.WithLock(func(s *store.Store) (any, error) {
+		return s.GetEntry(bytes(templatesKey))
+	})
+	if err != nil {
+		return []Template{}, err
```
yuce, line 219:
```suggestion
		return nil, err
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	err = json.Unmarshal(b.([]byte), &templates)
+	if err != nil {
+		return []Template{}, err
```
yuce, line 223:
```suggestion
		return nil, err
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	return templates, nil
+}
+
+func bytes(s string) []byte {
+	return []byte(s)
+}
```
yuce, line 230: `[]byte(s)` is quite idiomatic, so I think `bytes` function is not necessary.

yuce (changes requested):
Some ideas...

base/commands/project/project_list_it_test.go
```diff
@@ -0,0 +1,67 @@
...
+		sPath := filepath.Join(tcx.HomePath(), storeFolder)
+		sa := store.NewStoreAccessor(sPath, log.NopLogger{})
+		check.MustValue(sa.WithLock(func(s *store.Store) (any, error) {
+			err := s.SetEntry(bytes(nextFetchTimeKey),
+				bytes(strconv.FormatInt(time.Now().Add(cacheRefreshInterval).Unix(), 10)))
+			return nil, err
+		}))
+		check.MustValue(sa.WithLock(func(s *store.Store) (any, error) {
+			b, err := json.Marshal([]Template{{Name: "test_template"}})
+			check.Must(err)
```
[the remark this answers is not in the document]
kutluhanmetin: I did it on purpose that I need `b` on line 45: `err = s.SetEntry(bytes(templatesKey), b)`

base/commands/project/project_list_templates.go
kutluhanmetin: I think we should not use the word `can` that it is not related to ability. We are checking if we need to fetch the template list from remote or not.

base/commands/project/project_list_templates.go
kutluhanmetin: Let me check if the API provides paging 👍

kutluhanmetin: Yes, it is possible via API: https://api.github.com/users/hazelcast-templates/repos?per_page=10&page=1, but how should be our user experience to switch to the next page in CLC?

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,228 @@
...
+
+const (
+	flagLocal            = "local"
+	flagForce            = "force"
+	nextFetchTimeKey     = "project.templates.nextFetchTime"
+	templatesKey         = "project.templates"
+	cacheRefreshInterval = 10 * time.Minute
+)
+
+var storeFolder = filepath.Join("caches", "templates")
```
yuce, line 36: Could you add the following function in `paths.go`
```

func Caches() string {
	return filepath.Join(Home(), "caches")
}
```
I think with that change we can remove `storeFolder`.

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,228 @@
...
+	cc.AddBoolFlag(flagLocal, "", false, false, "list the templates which exist on local environment")
+	help := "Lists templates that can be used while creating projects."
+	cc.SetCommandHelp(help, help)
+	return nil
+}
+
+func (lc ListCmd) Exec(ctx context.Context, ec plug.ExecContext) error {
+	isLocal := ec.Props().GetBool(flagLocal)
+	isForce := ec.Props().GetBool(flagForce)
+	if isLocal && isForce {
```
yuce, line 56:
```suggestion
	local := ec.Props().GetBool(flagLocal)
	refresh := ec.Props().GetBool(flagForce)
	if local && refresh {
```

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,228 @@
...
+		fetch = true
+	}
+	if fetch || isForce {
+		ts, err := fetchTemplates()
+		if err != nil {
+			return nil, err
+		}
+		err = updateCache(sa, ts)
+		if err != nil {
+			logger.Debugf("Updating templates cache: %w", err)
```
yuce, line 108: Just to communicate that this is an error:
```suggestion
			logger.Debugf("Error: updating templates cache: %w", err)
```

base/commands/project/project_list_templates.go
yuce: How about `shouldFetch` ?

base/commands/project/project_list_templates.go
```diff
@@ -0,0 +1,228 @@
...
+
+func listTemplates(logger log.Logger, isLocal bool, isForce bool) ([]Template, error) {
+	sa := store.NewStoreAccessor(paths.Join(paths.Home(), storeFolder), logger)
+	if isLocal {
+		return listLocalTemplates()
+	}
+	var fetch bool
+	var err error
+	if fetch, err = isFetch(sa); err != nil {
+		logger.Debugf("Decide to fetch templates: %w", err)
```
yuce, line 97:
```suggestion
		logger.Debugf("Error: checking template list expiry: %w", err)
```

yuce (changes requested):
A few more ideas...

yuce (approved):
Looks good!

PR #339, opened 2023-08-11
Fixed unmarshalling error when IP Allowlist is configured on viridian

pmcgleenon:
Here's the log message from `clc viridian list-clusters` when a cluster has an IP Allow list configured

```
2023-08-11T19:47:32.962+0100    ERROR   viridian/common.go:228  listing clusters: json: cannot unmarshal array into Go struct field Cluster.Content.allowedIps of type string
```

Here's an example of what the IP looks like from the Viridian `/cluster/` endpoint - grabbed from the Viridian UI
{id: 177, ip: "207.154.227.247", description: null}


yuce (approved):
Looks good!

PR #353, opened 2023-08-23
[CLC-278] Fix CLC prompt after running config wizard

mtyazici:


clc/cmd/exec_context.go
```diff
@@ -107,6 +107,10 @@ func (ec *ExecContext) Props() plug.ReadOnlyProperties {
 	return ec.props
 }
 
+func (ec *ExecContext) ConfigPath() string {
```
yuce, line 110: Does this have to be in `ExecContext`?

clc/config/wizard_provider.go
```diff
@@ -13,40 +13,40 @@ import (
...
-	"github.com/hazelcast/hazelcast-commandline-client/clc/config"
+	"github.com/hazelcast/hazelcast-commandline-client/clc/config/wizard"
 	"github.com/hazelcast/hazelcast-commandline-client/clc/paths"
 	clcerrors "github.com/hazelcast/hazelcast-commandline-client/errors"
 	"github.com/hazelcast/hazelcast-commandline-client/internal/plug"
 )
 
-type Provider struct {
-	fp  *atomic.Pointer[config.FileProvider]
+type WizardProvider struct {
```
yuce, line 22: Is it necessary to move this provider from `clc/config/wizard`?

yuce (changes requested):
Some ideas...

clc/config/wizard_provider.go
mtyazici: It is not necessary but I believe it is simpler to have all providers in the same package scope.

clc/cmd/exec_context.go
mtyazici: I could not find anywhere else to put it since it depends on the Config Provider. Do you have a suggestion where I could put it?

yuce: OK, let's keep it for now.
We may need to add a method that returns the config provider in the future.

yuce (dismissed):
Looks good!

yuce (approved):

PR #360, opened 2023-08-31
Fix install script text

yuce:


yuce:
Trivial, merging it myself

PR #370, opened 2023-09-08
[CLC-312]: Add "cancel" command to DMT

kutluhanmetin:


base/commands/migration/cancel_stages.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (st *CancelStages) cancelStage(ctx context.Context) func(stage.Statuser) error {
+	return func(statuser stage.Statuser) error {
+		c := CancelItem{ID: st.migrationID}
+		b, err := json.Marshal(c)
+		if err != nil {
+			return err
+		}
+		st.cancelQueue.Put(ctx, serialization.JSON(b))
+		return err
```
srknzl, line 80:
```suggestion
		return nil
```

srknzl:
@kutluhanmetin Note: We will revert back to IQueue usage instead of ITopic for cancel queue. FYI. So you should update PR as well. See https://github.com/hazelcast/hazelcast-enterprise/pull/6451#discussion_r1319510673

base/commands/migration/cancel_stages.go
```diff
@@ -0,0 +1,106 @@
...
+
+func waitForCancel(ctx context.Context, ci *hazelcast.ClientInternal, migrationID string) error {
+	for {
+		s, err := fetchMigrationStatus(ctx, ci, migrationID)
+		if err != nil {
+			return err
+		}
+		if Status(s) == StatusCanceling {
+			return nil
+		}
```
srknzl, line 100: any sleep here?

kutluhanmetin: I remember we had a similar discussion for `WaitForMigrationToBeInProgress` and decided not to sleep. I think it is a similar situation in here as well.

srknzl (approved):

yuce (approved):

PR #372, opened 2023-09-08
Refactor integrate stage 2

yuce:


base/commands/list/list_set.go
```diff
@@ -6,54 +6,65 @@ import (
...
+		}
+		data := codec.DecodeListSetResponse(resp)
+		return convertDataToRow(ci, data, ec.Props().GetBool(base.FlagShowType))
 	})
 	if err != nil {
 		return err
 	}
 	stop()
-	return nil
+	return ec.AddOutputRows(ctx, rowV.(output.Row))
```
mtyazici, line 67: We are printing the last set value as follows
```
Value 
last-val  
```
We are also using the same title for every return type. I find this a bit confusing, what do you think?

yuce: > same title for every return type

Do you mean every data structure?

mtyazici: Not every data structure, I realized this function `convertDataToRow(ci *hazelcast.ClientInternal, data hazelcast.Data, showType bool)` is only used once in `list set`. 

What I meant is we are printing the correct title for every other command and for this one we are printing the last value with the `Value` title which is a bit misleading.

yuce: Changed the titles at:
* https://github.com/hazelcast/hazelcast-commandline-client/pull/372/commits/e9cfedad237d7557064057f20e20803dad14f0c9
*  https://github.com/hazelcast/hazelcast-commandline-client/pull/372/commits/9965a3088c4445fbc02d47e620525d8a0eb320c3

mtyazici (approved):

PR #386, opened 2023-09-13
[CLC-334] Fix list add command with index

mtyazici:


base/commands/list/list_add.go
```diff
@@ -65,7 +65,11 @@ func (AddCommand) Exec(ctx context.Context, ec plug.ExecContext) error {
 		if err != nil {
 			return nil, err
 		}
-		return codec.DecodeListAddResponse(resp), err
+		if index >= 0 {
+			return true, err
+		} else {
```
yuce, line 70: `else` is not necessary

yuce (changes requested):
Just a trivial idea...

yuce (approved):
Looks good!

PR #390, opened 2023-09-15
Fix project command tags

yuce:


yuce:
Trivial, merging it myself

PR #413, opened 2023-10-23
Build dmt tool for all supported platforms [HZ-2594]

frant-hartm:


yuce (changes requested):
@frant-hartm Would it make sense to add something like `release_dmt.sh` instead of updating `build_dmt.sh`? Since `build_dmt.sh` is frequently used during development/testing.

yuce (approved):

kutluhanmetin (approved):

PR #420, opened 2023-10-31
[CLC-425]: Print Warnings Before Exit

kutluhanmetin:


base/commands/migration/migration_stages.go
```diff
@@ -269,6 +269,22 @@ func finalizeMigration(ctx context.Context, ec plug.ExecContext, ci *hazelcast.C
...
+		ec.Logger().Error(err)
+		return
+	}
+	var warnings []string
+	err = json.Unmarshal(row.(serialization.JSON), &warnings)
+	if err != nil {
+		ec.Logger().Error(err)
+		return
+	}
+	ec.PrintlnUnnecessary("* " + strings.Join(warnings, "\n* "))
```
srknzl, line 285: If there're lots of them, it can be more reasonable to point to the migration report instead. The warnings will be in the report as well. If there're less than 5 we can directly print them

kutluhanmetin: I think there won't be too many warning that makes it hard to display in the terminal.

srknzl: There can be in the future. We'll keep adding things to warnings field.

kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/420/commits/64e39bc69b8d0d2ef7548d97ea3f9308c64ab314

srknzl: Note that, it's not just for future. 1000 ds skips can even happen now.

base/commands/migration/migration_stages.go
```diff
@@ -269,6 +269,24 @@ func finalizeMigration(ctx context.Context, ec plug.ExecContext, ci *hazelcast.C
...
+	}
+	var warnings []string
+	err = json.Unmarshal(row.(serialization.JSON), &warnings)
+	if err != nil {
+		ec.Logger().Error(err)
+		return
+	}
+	if len(warnings) <= 5 {
+		ec.PrintlnUnnecessary("* " + strings.Join(warnings, "\n* "))
+	}
```
srknzl, line 287: Can we add a else block that says "you have x warnings that you can find in your migration report" ?

kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/420/commits/2147b151d2b68f986efa9cac72febaaa27e80b7d

yuce (approved):

PR #423, opened 2023-11-01
[CLC-426]: Show Overall Remaining Time and Completion Percentage

kutluhanmetin:


base/commands/migration/migration_stages.go
```diff
@@ -233,6 +241,46 @@ func fetchMigrationStatus(ctx context.Context, ci *hazelcast.ClientInternal, mig
...
+			return 0, 0, err
+		}
+		// convert it to parsable duration format
+		rtStr := fmt.Sprintf("%sms", remainingTime.(serialization.JSON).String())
+		rt, err := time.ParseDuration(rtStr)
+		if err != nil {
+			return 0, 0, err
+		}
+		cpStr := completionPercentage.(serialization.JSON).String()
+		cp, err := strconv.ParseFloat(cpStr, 32)
```
srknzl, line 275: why not 64?

srknzl (commented):
can't we test this change 🤔  ?

base/commands/migration/migration_stages.go
kutluhanmetin: Statuser's `SetProgress` method accepts `float32` values.

kutluhanmetin:
> can't we test this change 🤔 ?

I think it is a bit hard to test it, because values are changing constantly. And it is very easy to detect with manual tests that it wouldn't worth the effort IMO.

srknzl (approved):

srknzl:
I was talking about maybe running dmt estimate and checking the output has the "remaining time x seconds" text. @kutluhanmetin 

but it's not too important if it's difficult

base/commands/migration/migration_stages.go
```diff
@@ -76,6 +77,13 @@ func createMigrationStages(ctx context.Context, ec plug.ExecContext, ci *hazelca
 					case StatusCanceled, StatusCanceling:
 						execErr = clcerrors.ErrUserCancelled
 						break statusReaderLoop
+					case StatusInProgress:
+						rt, cp, err := fetchOverallProgress(ctx, ci, migrationID)
+						if err != nil {
+							ec.Logger().Error(err)
+						}
+						status.SetProgress(cp)
+						status.SetRemainingDuration(rt)
```
yuce, line 86: If there's an error the progress and the remaining duration would be set to 0.
Is that the expected outcome?

base/commands/migration/migration_stages.go
```diff
@@ -233,6 +241,46 @@ func fetchMigrationStatus(ctx context.Context, ci *hazelcast.ClientInternal, mig
...
+func fetchOverallProgress(ctx context.Context, ci *hazelcast.ClientInternal, migrationID string) (time.Duration, float32, error) {
+	q := fmt.Sprintf(`SELECT JSON_QUERY(this, '$.remainingTime'), JSON_QUERY(this, '$.completionPercentage') FROM %s WHERE __key='%s'`, StatusMapName, migrationID)
+	res, err := ci.Client().SQL().Execute(ctx, q)
+	if err != nil {
+		return 0, 0, err
+	}
+	it, err := res.Iterator()
+	if err != nil {
+		return 0, 0, err
+	}
```
yuce, line 253: This can be refactored to a function.

base/commands/migration/migration_stages.go
```diff
@@ -233,6 +241,46 @@ func fetchMigrationStatus(ctx context.Context, ci *hazelcast.ClientInternal, mig
...
+	if it.HasNext() {
+		// single iteration is enough that we are reading single result for a single migration
+		row, err := it.Next()
+		if err != nil {
+			return 0, 0, err
+		}
```
yuce, line 259: I thought you've introduced a function to query SQL and return a single result

base/commands/migration/migration_stages.go
```diff
@@ -233,6 +241,46 @@ func fetchMigrationStatus(ctx context.Context, ci *hazelcast.ClientInternal, mig
...
+		completionPercentage, err := row.Get(1)
+		if err != nil {
+			return 0, 0, err
+		}
+		// convert it to parsable duration format
+		rtStr := fmt.Sprintf("%sms", remainingTime.(serialization.JSON).String())
+		rt, err := time.ParseDuration(rtStr)
+		if err != nil {
+			return 0, 0, err
+		}
```
yuce, line 273: Why not just convert to integer and multiply by `time,Millisecond`?

yuce (changes requested):
A few ideas...

base/commands/migration/migration_stages.go
kutluhanmetin: Shall we set the text if there is an error and not set progress and duration:
```
status.SetText("Unable to calculate remaining duration and progress")
```

base/commands/migration/migration_stages.go
kutluhanmetin: You are referring to querySingleRow that only uses `row.Get(0)` to fetch single result. But in here I have two values to read that I could not reuse it.

base/commands/migration/migration_stages.go
kutluhanmetin: I think it is should be ok like that. It is not reused.

base/commands/migration/migration_stages.go
kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/423/commits/262b52601907752ed8c03cfedea1108e43aa6a2c

base/commands/migration/migration_stages.go
srknzl: If we can't fetch the progress, I would not update them.

yuce: Yeah, maybe something like that.

base/commands/migration/migration_stages.go
yuce: Maybe `querySingleRow` could be changed to return the row, not the first column.

kutluhanmetin: The other functions use that function in common, so it is re-used a couple of times. Shall we keep it as is?

base/commands/migration/migration_stages.go
kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/423/commits/9a153d5ad95ee2477f02f35194241ad60932afb4

base/commands/migration/migration_stages.go
```diff
@@ -76,6 +77,15 @@ func createMigrationStages(ctx context.Context, ec plug.ExecContext, ci *hazelca
...
 						execErr = clcerrors.ErrUserCancelled
 						break statusReaderLoop
+					case StatusInProgress:
+						rt, cp, err := fetchOverallProgress(ctx, ci, migrationID)
+						if err != nil {
+							ec.Logger().Error(err)
+							status.SetText("Unable to calculate remaining duration and progress")
+						} else {
+							status.SetProgress(cp)
+							status.SetRemainingDuration(rt)
```
yuce, line 87: What happens if the progress status is received after setting the ""Unable to calculate remaining duration and progress" text? Is it changed to "XXX is in progress?

base/commands/migration/migration_stages.go
yuce: If it's not too hard to make that change, I would change function I've described and change the code where it is called.

kutluhanmetin: I didn't mean to say it is hard, I think it is not very useful.

base/commands/migration/migration_stages.go
kutluhanmetin: It is a bit hard to run this scenario on my local environment. I was expecting it to be updated with the stage's ProgressMsg. Is it wrong?

yuce: No, progress message is set once, before the stage starts executing, see: clc/ux/stage/stage.go:152

base/commands/migration/migration_stages.go
yuce: There are only 4 places `querySingleRow` is used, so not much code will be changed with that refactoring, and it will be more generic, so it can also be used ^.
You can also change its return value to return `[]any` to return all column values.

base/commands/migration/migration_stages.go
kutluhanmetin: ok, this commit fixes it: https://github.com/hazelcast/hazelcast-commandline-client/pull/423/commits/465a2e06ed87b8183027d0514261a31517b2dd2e

base/commands/migration/migration_stages.go
kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/423/commits/c53f607d67077525ef6a3b9e419dd990fc635f14

yuce (approved):

PR #424, opened 2023-11-02
[CLC-427][CLC-428]: Remove Data Migrations in Progress List and Status Map Usages

kutluhanmetin:
* Remove Data Migrations in Progress List and Status Map Usages
* Refactor test files
* Add new stage "prechecks" to start command
* Wait for migration to be in STARTED status in start command


base/commands/migration/cancel_it_test.go
```diff
@@ -34,30 +35,36 @@ func noMigrationsCancelTest(t *testing.T) {
...
-		tcx.WithReset(func() {
+		setStatusInProgress(tcx, ctx)
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go tcx.WithReset(func() {
+			defer wg.Done()
 			Must(tcx.CLC().Execute(ctx, "cancel"))
 		})
-		MustValue(l.Remove(ctx, serialization.JSON(m)))
+		time.Sleep(1 * time.Second)
```
yuce, line 57: We may have a problem if these tests don't work without sleeping.

yuce, line 57: We may have a problem if these tests don't work without sleeping.

kutluhanmetin: I could not find a way to work them without sleeping, because it is a simulation of changing states. State changes does not depend on any condition that eventually does not help. So we cannot simulate the behaviour in any other way. I debugged these tests and without sleeping it immediately changes the state to latest state that commands cannot process the initial states which makes the tests fail. Sorry, I looked for it for a while and I have no other solution for this. Also I think we don't introduce flaky tests, because 1 second it a good enough time a command to process the initial state on any machine IMO.

srknzl: (also I don't get why we have sleep in this test, how mocking of migration cluster is done is complex for me. )

kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/424/commits/0ba87e67479948d8b7dff6fd94a9a179d50eed57

kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/424/commits/0ba87e67479948d8b7dff6fd94a9a179d50eed57

yuce:
@kutluhanmetin The tests fail. Could you have a look?

srknzl:
I have tried this version with the latest backend in the PR https://github.com/hazelcast/hazelcast-enterprise/pull/6765

The result of a dmt cancel when there's no migration in progress: 

image

I guess this error is not expected

srknzl:
Status command when there's no migration: 

➜  hazelcast-enterprise-5.3.5-DM-SNAPSHOT-slim ./dmt -c migration.yaml status

Hazelcast Data Migration Tool v5.3.0
(c) 2023 Hazelcast, Inc.

 ERROR Could not connect to the migration cluster: finding migration in progress: no rows found

srknzl:
when there was a migration and I canceled while it's in progress: 

image


This worked correctly except that the start command exited without any message. We should fix that as well (not related to this PR)

srknzl (approved):

srknzl:
Note: I approved because I retried all the scenarios above and all of them is fixed now.

yuce (approved):

PR #428, opened 2023-11-07
[CLC-432]: Fix hanging problem when source and target clusters is not running

kutluhanmetin:
- The reason of the problem is that we forgot to handle context errors in for loop while waiting migration to be in progress, so even if we have a timeout error it waits for it to be in progress forever.
- Also this PR increases the timeout of the context, because it was 30 seconds which is same with the backend. So sometimes it used to fail with timeout, sometimes with `Unable to connect to any cluster` error. After increasing the timeout, user will consistently see the `Unable to connect to any cluster` error.


srknzl (approved):
seems correct, needs testing. left to qa

yuce (approved):

PR #429, opened 2023-11-07
[CLC-431]: dmt command hangs when unable to connect w/ the migration cluster

kutluhanmetin:


srknzl (commented):
Wont the client hang in the connect stage now? Don't we need to add some timeout ? 

EDIT: I guess we won't do that due to the timeout flag in clc, but do we document that in the help (-h) page? Also we should have the timeout by default IMO.

srknzl:
Also do we print what @yuce suggested here: https://hazelcast.atlassian.net/browse/CLC-431?focusedCommentId=88217 ?

kutluhanmetin:
@srknzl yes, we are printing what @yuce suggested, it will hang with “Connecting …”.

srknzl (approved):

PR #436, opened 2023-11-14
[CLC-442]: Fix defers

kutluhanmetin:


base/commands/migration/migration_status.go
```diff
@@ -33,10 +33,12 @@ func (s StatusCmd) Exec(ctx context.Context, ec plug.ExecContext) (err error) {
 		return err
 	}
 	defer func() {
-		maybePrintWarnings(ctx, ec, sts.ci, mID.(string))
-		finalizeErr := finalizeMigration(ctx, ec, sts.ci, mID.(string), ec.Props().GetString(flagOutputDir))
-		if err == nil {
-			err = finalizeErr
+		if sts.ci != nil {
+			maybePrintWarnings(ctx, ec, sts.ci, mID.(string))
+			finalizeErr := finalizeMigration(ctx, ec, sts.ci, mID.(string), ec.Props().GetString(flagOutputDir))
```
yuce, line 38: Could you change the `stage.Execute` call above so it uses `""` instead of `any(nil)` as the initial state value?
```
mID, err := stage.Execute(ctx, ec, "", sp)
```
So the state type is `string` and casting it to `string` here is not necessary.

yuce, line 38: Could you change the stage.Execute call above so it uses `""` instead of `any(nil)` as the initial state value?
```
mID, err := stage.Execute(ctx, ec, "", sp)
```
So the state type is string and casting it to string here is not necessary.

yuce (changes requested):

srknzl (approved):

base/commands/migration/migration_status.go
kutluhanmetin: I changed it that way, but it still requires type assertion that it gives the following compilation error:
```
Error: base/commands/migration/migration_status.go:36:40: cannot use mID (variable of type any) as string value in argument to maybePrintWarnings: need type assertion
```

kutluhanmetin: https://github.com/hazelcast/hazelcast-commandline-client/pull/436/commits/b439ab9af483ad11c50b16aa8d97109355a3d725
https://github.com/hazelcast/hazelcast-commandline-client/pull/436/commits/a4580f4d79244179436e7f83c127726907706e49

yuce (approved):

